PR #24, opened 2020-01-07
feat: add enforce linkage checker

athakor:
Towards [#6396](https://github.com/googleapis/google-cloud-java/issues/6396)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,14 +63,14 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <github.global.server>github</github.global.server>
     <site.installationModule>google-cloud-bigquerystorage-parent</site.installationModule>
-    <google.core.version>1.91.1</google.core.version>
+    <google.core.version>1.92.1</google.core.version>
     <google.api-common.version>1.8.1</google.api-common.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -173,6 +178,30 @@
...
+                  </requireMavenVersion>
+                  <requireJavaVersion>
+                    <version>[1.7,)</version>
+                  </requireJavaVersion>
+                  <requireUpperBoundDeps/>
+                </rules>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
```
chingor13: We already include this check in the shared-config parent project

athakor: thanks,for the info.so can we add one more rule over there for linkage checker ?
```
<LinkageCheckerRule implementation="com.google.cloud.tools.dependencies.enforcer.LinkageCheckerRule">
  <reportOnlyReachable>true</reportOnlyReachable>
</LinkageCheckerRule>
```

elharo: I don't think you want `<reportOnlyReachable>true</reportOnlyReachable>` here.

athakor:
This enforcer config is already present in the shared-config since at least version 0.3.0 as can be seen [here](https://github.com/googleapis/java-shared-config/blob/v0.3.0/pom.xml#L204-L226)

PR #53, opened 2020-02-03
deps: bump dependency versions

stephaniewang526:
Fixes #31 #51 #52

### DIFF
--- a/google-cloud-bigquerystorage/pom.xml
+++ b/google-cloud-bigquerystorage/pom.xml
@@ -105,7 +105,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-bigquery</artifactId>
-      <version>1.102.0</version>
+      <version>1.106.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/pom.xml
```diff
@@ -105,7 +105,7 @@
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-bigquery</artifactId>
-      <version>1.102.0</version>
+      <version>1.106.0</version>
```
elharo: This is likely the upgrade that's causing the issues. I'd try splitting this out into a separate PR to isolate it.

pom.xml
```diff
@@ -108,6 +110,16 @@
         <artifactId>grpc-google-cloud-bigquerystorage-v1beta2</artifactId>
         <version>0.85.2-SNAPSHOT</version><!-- {x-version-update:grpc-google-cloud-bigquerystorage-v1beta2:current} -->
       </dependency>
+      <dependency>
```
elharo: I don't think you need to add these two. Just move avro to the end of your dependencies section.

stephaniewang526: will do

elharo: Or maybe not. The avro dependency on jackson is not as deep.

PR #117, opened 2020-03-23
feat: add flatten plugin

stephaniewang526:
add flatten plugin to flatten pom.xml


pom.xml
```diff
@@ -77,6 +77,7 @@
     <threeten.version>1.4.2</threeten.version>
     <javax.annotations.version>1.3.2</javax.annotations.version>
     <animal-sniffer.version>1.18</animal-sniffer.version>
+    <opencensus-api.version>0.25.0</opencensus-api.version>
```
elharo, line 80: 0.24.0 is what we're sticking to until gRPC updates

pom.xml
```diff
@@ -250,6 +327,7 @@
     <module>grpc-google-cloud-bigquerystorage-v1</module>
     <module>google-cloud-bigquerystorage</module>
     <module>google-cloud-bigquerystorage-bom</module>
+    <module>samples</module>
```
chingor13, line 330: We don't want this in the main `<modules>` list -- we don't want others to have to run the samples tests to contribute to the main artifacts. Instead, we'll put it in the `enable-samples` profile like: https://github.com/googleapis/java-logging/blob/6f4fae2141662f0421364ea25f738366f3feaf41/pom.xml#L312-L319

PR #141, opened 2020-04-04
fix: waitForTermination in the manual client #140

### LINKED ISSUE #140: Stream is not properly closed in writeapi manual client
We need to close the stream in shutdown.

yirutang:
Fixes #140 ☕️

After the fix, close() should fully work. So I updated the unit test to use close().
Also fixed a issue in reconnection.


stephaniewang526:
Please correct initial commit msg to "fix: waitForTermination" so that conventionalcommit check will pass. FYI, this is also used in triggering our release pipeline.

PR #163, opened 2020-04-09
ci: add dependency list completeness check

stephaniewang526:


.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,50 @@
...
+set -eo pipefail
+
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+function now() { date +"%Y-%m-%d %H:%M:%S" | tr -d '\n'; }
+function msg() { println "$*" >&2; }
+function println() { printf '%s\n' "$(now) $*"; }
+
+function cleanup() {
+  rm .org-list.txt .new-list.txt .diff.txt
```
chingor13, line 25: We should consider not deleting these and even preserving the as artifacts of the build - it's possible to store them so you don't need to look through the log.

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
chingor13, line 1: It looks like you might also need to make the file permissions executable.

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,50 @@
...
+msg "Generating dependency list using original pom..."
+mvn dependency:list -f pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:compile' >.org-list.txt
+
+# Output dep list with compile scope generated using the flattened pom
+msg "Generating dependency list using flattened pom..."
+mvn dependency:list -f .flattened-pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:compile' >.new-list.txt
+
+# Compare two dependency lists
+msg "Comparing dependency lists..."
+diff .org-list.txt .new-list.txt >.diff.txt && (msg "Success. No diff!" && cleanup) || (msg "Diff found: " && cat .diff.txt)
```
chingor13, line 50: This looks like it will return a 0 exit code even if it fails because the second part of the command returns 0

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+## Locally install artifacts
+#mvn verify -DskipTests
+
+## Find client directory that continas flattened pom and cd into it
+clientDir=$(find -iname ".flattened-pom.xml" | cut -d"/" -f1-2)
```
chingor13, line 35: If we plan to templatize this file and roll it out to all clients, we should be able to skip this if flattening is not yet enabled (without failing the build). Since it's possible that we might flatten multiple artifacts, this could be turned into something like:
```
for flattened_pom in `find -name ".flattened-pom.xml"`
do
  # check each ${flattened_pom}
done
```

Then if there are no flattened poms, nothing happens and the check passes.

.kokoro/completeness-check.sh
stephaniewang526: I will not remove them. Please let me know if there's a specific way you would like to store them.

.kokoro/completeness-check.sh
stephaniewang526: this file is alr an executable.

.kokoro/completeness-check.sh
stephaniewang526: Done

.kokoro/completeness-check.sh
stephaniewang526: Done.

.kokoro/dependencies.sh
```diff
@@ -36,3 +36,16 @@ retry_with_backoff 3 10 \
     -Dclirr.skip=true
 
 mvn -B dependency:analyze -DfailOnWarning=true
+
+echo "****************** DEPENDENCY LIST COMPLETENESS CHECK *******************"
+source ${scriptDir}/completeness-check.sh
+
+## Locally install artifacts
+mvn verify -DskipTests
```
chingor13, line 44: We shouldn't need this line, it's being installed on line 33

.kokoro/completeness-check.sh
```diff
@@ -13,38 +13,31 @@
...
-#mvn verify -DskipTests
-
-## Find client directory that continas flattened pom and cd into it
-clientDir=$(find -iname ".flattened-pom.xml" | cut -d"/" -f1-2)
-cd "$clientDir"
-
 ## Run dependency list completeness check
+function completenessCheck() {
+  # cd into dir containing flattened pom
+  cd "$1" || exit
```
chingor13, line 26: If you're planning on changing directories, then back, I prefer `pushd` then `popd` to go back the last directory.

chingor13: Also, I think `exit` will completely exit the script. You probably want `return <some non-zero code>` which can be dealt with by the caller.

stephaniewang526: make sense!

.kokoro/dependencies.sh
stephaniewang526: Done

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,43 @@
...
+  msg "Generating dependency list using original pom..."
+  mvn dependency:list -f pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:compile' >.org-list.txt
+
+  # Output dep list generated using the flattened pom (test scope deps are ommitted)
+  msg "Generating dependency list using flattened pom..."
+  mvn dependency:list -f .flattened-pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' >.new-list.txt
+
+  # Compare two dependency lists
+  msg "Comparing dependency lists..."
+  diff .org-list.txt .new-list.txt >.diff.txt && msg "Success. No diff!" && return 0|| msg "Diff found. Check .diff.txt file located in $1. Exiting with exit code $?." && return 1
```
chingor13, line 39: If you do return here, the directory is not restored.

.kokoro/dependencies.sh
```diff
@@ -27,6 +27,16 @@ source ${scriptDir}/common.sh
 java -version
 echo $JOB_TYPE
 
+echo "****************** DEPENDENCY LIST COMPLETENESS CHECK *******************"
```
chingor13, line 30: This needs to run after the `mvn install` otherwise there won't be a flattened pom.xml

stephaniewang526: ah that's correct.

.kokoro/dependencies.sh
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eo pipefail
+set -o pipefail
```
chingor13, line 16: this would allow `mvn -B dependency:analyze -DfailOnWarning=true` to not fail the build if it finds issues

stephaniewang526: ok -- I put it back

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+#set -eo pipefail
+
+## Helper functions
+function now() { date +"%Y-%m-%d %H:%M:%S" | tr -d '\n'; }
+function msg() { println "$*" >&2; }
+function println() { printf '%s\n' "$(now) $*"; }
+
+## Run dependency list completeness check
```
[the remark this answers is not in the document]
stephaniewang526: okay - i will remove the line that's commented out.

.kokoro/completeness-check.sh
```diff
@@ -0,0 +1,43 @@
...
+function completenessCheck() {
+  # Output dep list with compile scope generated using the original pom
+  msg "Generating dependency list using original pom..."
+  mvn dependency:list -f pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' | grep -v ':test$' >.org-list.txt || true # continue without error if list is empty
```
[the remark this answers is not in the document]
stephaniewang526: yes agree

PR #174, opened 2020-04-15
fix: custom retry logic

### LINKED ISSUE #86: [bigquerystorage, Java] Reconnect on "EOS on DATA frame" Internal Server Errors
original ticket: b/146489609

#### copy from bug
The Java client needs to reconnect to the BigQuery Storage stream when an "EOS on DATA frame" error occurs.

See: https://groups.google.com/a/google.com/d/topic/cloud-dpes-bigquery/dzvlD5O00a0/discussion for discussion.

Implementation in Python: https://github.com/googleapis/google-cloud-python/pull/9994

pmakani:

Fixes #86

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/stub/readrows/BigQueryStorageException.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/stub/readrows/BigQueryStorageException.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery.storage.v1.stub.readrows;
+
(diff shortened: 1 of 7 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/stub/readrows/BigQueryStorageException.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery.storage.v1.stub.readrows;
+
+import com.google.cloud.grpc.BaseGrpcServiceException;
+
+public class BigQueryStorageException extends BaseGrpcServiceException {
```
kmjung, line 21: Am I right in understanding that exceptions which are visible to end users today are all BaseGrpcServiceExceptions, and so this should prevent this change from being a breaking change from a user perspective? I'm wondering about the case where a caller has already implemented their own special-case handling for e.g. exceptions produced on INTERNAL error responses from the server.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/stub/readrows/BigQueryStorageExceptionFactory.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  public static BigQueryStorageException newBigQueryStorageException(Throwable throwable) {
+    Status status = Status.fromThrowable(throwable);
+    return new BigQueryStorageException(
+        status.getDescription(),
+        throwable.getCause(),
+        status.getCode().value(),
+        isRetryable(ErrorCode.fromGrpcStatus(status), throwable.getCause()));
```
kmjung, line 113: Is this sufficient to cause the client to retry on failure? I don't see a test which verifies the retry behavior.

PR #186, opened 2020-04-17
feat: use shared-dependencies bom and add flatten to grpc and proto m…

stephaniewang526:
1. use shared-deps instead of a bunch of deps
2. add flatten plugin to grpc and proto modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,19 +64,7 @@
     <project.protobuf-java.version>3.11.4</project.protobuf-java.version>
     <github.global.server>github</github.global.server>
     <site.installationModule>google-cloud-bigquerystorage-parent</site.installationModule>
-    <google.core.version>1.93.4</google.core.version>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -141,93 +140,21 @@
...
-        <groupId>javax.annotation</groupId>
-        <artifactId>javax.annotation-api</artifactId>
-        <version>${javax.annotations.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>animal-sniffer-annotations</artifactId>
-        <version>${animal-sniffer.version}</version>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
```
chingor13, line 150: This is probably a candidate to add to the shared dependencies bom

stephaniewang526: agree - pls let me know when you add

PR #201, opened 2020-04-21
chore: add temp code replacement hack due to gapic code-gen template bug

stephaniewang526:
Depends on https://github.com/googleapis/gapic-generator/issues/3181

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -34,7 +34,24 @@ for version in versions:
       destination_name='bigquerystorage',
   )
 
-java.common_templates(excludes=[
-    #TODO: REMOVE THIS WHEN GRPC FIXES ITS TEMPLATE
-    'google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/MockBigQueryWriteImpl.java'
-])
+java.common_templates()
+
+# TODO: Remove all below s.replace() logic when upstream correction is made in gapic
+# Remove line added by gapic generator
+s.replace("google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/MockBigQueryWriteImpl.java",
+          "final Object response = responses.remove();",
+          "")
+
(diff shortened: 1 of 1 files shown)


synth.py
```diff
@@ -34,4 +34,7 @@
       destination_name='bigquerystorage',
   )
 
-java.common_templates()
+java.common_templates(excludes=[
+    #TODO: REMOVE THIS WHEN GRPC FIXES ITS TEMPLATE
+    'google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/MockBigQueryWriteImpl.java'
```
chingor13, line 39: This file is not a common template, so this won't work. The client copying happens in the `java.bazel_library` command above and does not include an option for excluding files.

To fix in synth, you will want to use `s.replace(files, search_regex, replace_string)`

stephaniewang526: Done.

synth.py
```diff
@@ -35,3 +35,23 @@
   )
 
 java.common_templates()
+
+# TODO: Remove all below s.replace() logic when upstream correction is made in gapic
```
chingor13, line 39: Please link to the upstream issue on gapic-generator

synth.py
```diff
@@ -35,3 +35,23 @@
...
+          public void onNext(AppendRowsRequest value) {
+            if (response instanceof AppendRowsResponse) {
+          """,
+          """
+          public void onNext(AppendRowsRequest value) {
+            requests.add(value);
+            final Object response = responses.remove();
+            if (response instanceof AppendRowsResponse) {
+          """
+          )
```
chingor13, line 57: You may want to re-run the java formatter after doing the code replacements.

synth.py
stephaniewang526: Done.

synth.py
stephaniewang526: Done.

PR #239, opened 2020-05-01
ci: temp fix to exclude build.bat before switching over to actions

stephaniewang526:


synth.py
```diff
@@ -35,27 +35,7 @@
...
-            requests.add(value);
-            final Object response = responses.remove();
-            if (response instanceof AppendRowsResponse) {
-          """
-          )
-
-# Re-run java code formatter after making hacky code change
-os.system("mvn com.coveo:fmt-maven-plugin:format")
\ No newline at end of file
+# TODO: remove this when we switch to actions
```
chingor13, line 38: nit: move this to the next line so it's clear that the exclusion is to be removed, not the entire common_templates call.

stephaniewang526: done

PR #260, opened 2020-05-07
dep: bump google-cloud-bigquery version to 1.113.0

stephaniewang526:


stephaniewang526:
it was skipped since 1.113.0 came out shortly after 1.112.0 (on the same day).

PR #328, opened 2020-06-05
Schema Compact

allenc3:
Initial work on Schema Compact. Almost finished with checking non supported types, just need to implement a cycle detection algorithm to prevent improper message nesting for protobufs.

Added some tests and proto files, and in order to check for annotations in SchemaCompact.java, I had to add a "compile" goal to protoc.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-bigquery</artifactId>
-        <version>1.116.1</version>
+        <version>1.116.2</version>
       </dependency>
       <dependency>
         <groupId>com.google.api.grpc</groupId>


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/SchemaCompact.java
```diff
@@ -96,4 +118,108 @@ public void check(String tableName, Descriptors.Descriptor userSchema)
...
+   * @param allMessageTypes Keeps track of all message types seen, and make sure they don't repeat
+   *     as recursive protos are not supported.
+   * @return True if field type and option is supported
+   * @throws IllegalArgumentException if schema is invalid
+   */
+  private static boolean isSupportedImpl(
+      Descriptors.Descriptor userSchema, HashSet<Descriptors.Descriptor> allMessageTypes)
+      throws IllegalArgumentException {
+
+    List<Descriptors.OneofDescriptor> oneofs = userSchema.getOneofs();
```
yirutang, line 175: Sorry, the doc is wrong. Oneof can be supported by flattening the fields out. Let's just leave it out of the check.

google-cloud-bigquerystorage/pom.xml
```diff
@@ -108,6 +108,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.zetasql</groupId>
```
stephaniewang526, line 112: is this a credibly sourced artifact? do we have a more well-known alternative to this?

synth.metadata
```diff
@@ -11,7 +11,7 @@
       "git": {
         "name": ".",
         "remote": "https://github.com/googleapis/java-bigquerystorage.git",
-        "sha": "98750d605dd7a6fead678a0b4573288ddd659277"
+        "sha": "f75eaafb3a20c541191686d1540248b67150709a"
```
stephaniewang526, line 14: these changes should not be rolled in from this feature PR.

.kokoro/build.sh
```diff
@@ -51,9 +51,7 @@ test)
     RETURN_CODE=$?
     ;;
 lint)
-    mvn \
-      -Penable-samples \
-      com.coveo:fmt-maven-plugin:check
+    mvn com.coveo:fmt-maven-plugin:check
```
stephaniewang526, line 54: this change should not be part of this PR

PR #342, opened 2020-06-11
fix: adding more logging to retry test

yirutang:
There is a test failure that we kept sending retry requests exceeding max attempt number.
Could be due to some race on currentRetries count? I couldn't repro it on my local machine.

Added some logging and moved currentRetries count after we set the exception.


stephaniewang526:
Please rebase this branch on master and only include the single new commit you want in the PR.

yirutang:
Rebase worked, PTAL.

PR #400, opened 2020-07-07
feat: Adding JsonToProtoMessage.java

allenc3:
Added JsonToProtoMessage.java which can convert json data to protocol buffer data given the protocol buffer descriptor.

### DIFF
--- a/google-cloud-bigquerystorage/pom.xml
+++ b/google-cloud-bigquerystorage/pom.xml
@@ -108,6 +108,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20200518</version>
+  </dependency>
+
 
     <!-- Test dependencies -->
     <dependency>
--- a/google-cloud-bigquerystorage/pom.xml
+++ b/google-cloud-bigquerystorage/pom.xml
@@ -111,7 +111,6 @@
     <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
-      <version>20200518</version>
   </dependency>
 
 
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
@@ -20,10 +20,8 @@ import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.Message;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
+import com.google.protobuf.UninitializedMessageException;
+import java.util.HashSet;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -76,69 +74,53 @@ public class JsonToProtoMessage {
       boolean topLevel,
       boolean allowUnknownFields)
       throws IllegalArgumentException {
-
     DynamicMessage.Builder protoMsg = DynamicMessage.newBuilder(protoSchema);
-    List<FieldDescriptor> protoFields = protoSchema.getFields();
-
-    Set<String> protoFieldNames = new TreeSet<String>(String.CASE_INSENSITIVE_ORDER);
-    for (FieldDescriptor field : protoFields) {
+    HashSet<String> protoFieldNames = new HashSet<String>();
+    for (FieldDescriptor field : protoSchema.getFields()) {
       protoFieldNames.add(field.getName());
     }
-
-    HashMap<String, String> jsonLowercaseNameToJsonName = new HashMap<String, String>();
     String[] jsonNames = JSONObject.getNames(json);
-    for (int i = 0; i < jsonNames.length; i++) {
-      jsonLowercaseNameToJsonName.put(jsonNames[i].toLowerCase(), jsonNames[i]);
-    }
-
-    if (!allowUnknownFields) {
-      for (int i = 0; i < jsonNames.length; i++) {
-        if (!protoFieldNames.contains(jsonNames[i])) {
-          throw new IllegalArgumentException(
-              String.format(
-                  "JSONObject has fields unknown to BigQuery: %s.%s. Set allowUnknownFields to True to allow unknown fields.",
-                  jsonScope, jsonNames[i]));
-        }
-      }
-    }
-
     int matchedFields = 0;
-    for (FieldDescriptor field : protoFields) {
-      String lowercaseFieldName = field.getName().toLowerCase();
-      String currentScope = jsonScope + "." + field.getName();
-
-      if (!jsonLowercaseNameToJsonName.containsKey(lowercaseFieldName)) {
-        if (field.isRequired()) {
-          throw new IllegalArgumentException(
-              "JSONObject does not have the required field " + currentScope + ".");
-        } else {
-          continue;
+    if (jsonNames != null) {
+      for (int i = 0; i < jsonNames.length; i++) {
+        String jsonName = jsonNames[i];
+        String jsonLowercaseName = jsonName.toLowerCase();
+        String currentScope = jsonScope + "." + jsonName;
+        if (!protoFieldNames.contains(jsonLowercaseName)) {
+          if (!allowUnknownFields) {
+            throw new IllegalArgumentException(
+                String.format(
+                    "JSONObject has fields unknown to BigQuery: %s. Set allowUnknownFields to True to allow unknown fields.",
+                    currentScope));
+          } else {
+            continue;
+          }
         }
-      }
-      matchedFields++;
-      if (!field.isRepeated()) {
-        fillField(
-            protoMsg,
-            field,
-            json,
-            jsonLowercaseNameToJsonName.get(lowercaseFieldName),
-            currentScope,
-            allowUnknownFields);
-      } else {
-        fillRepeatedField(
-            protoMsg,
-            field,
-            json,
-            jsonLowercaseNameToJsonName.get(lowercaseFieldName),
-            currentScope,
-            allowUnknownFields);
+        FieldDescriptor field = protoSchema.findFieldByName(jsonLowercaseName);
+        matchedFields++;
+        if (!field.isRepeated()) {
+          fillField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+        } else {
+          fillRepeatedField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+        }
       }
     }
     if (matchedFields == 0 && topLevel) {
       throw new IllegalArgumentException(
           "There are no matching fields found for the JSONObject and the protocol buffer descriptor.");
     }
-    return protoMsg.build();
+    DynamicMessage msg;
+    try {
+      msg = protoMsg.build();
+    } catch (UninitializedMessageException e) {
+      String errorMsg = e.getMessage();
+      int idxOfColon = errorMsg.indexOf(":");
+      String missingFieldName = errorMsg.substring(idxOfColon + 2);
+      throw new IllegalArgumentException(
+          String.format(
+              "JSONObject does not have the required field %s.%s.", jsonScope, missingFieldName));
+    }
+    return msg;
   }
 
   /**
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
@@ -52,6 +52,9 @@ public class JsonToProtoMessage {
   public static DynamicMessage convertJsonToProtoMessage(
       Descriptor protoSchema, JSONObject json, boolean allowUnknownFields)
       throws IllegalArgumentException {
+    if (json == null) {
+      throw new IllegalArgumentException("JSONObject is null.");
+    }
     if (json.length() == 0) {
       throw new IllegalArgumentException("JSONObject is empty.");
     }
@@ -79,32 +82,35 @@ public class JsonToProtoMessage {
 
     DynamicMessage.Builder protoMsg = DynamicMessage.newBuilder(protoSchema);
     String[] jsonNames = JSONObject.getNames(json);
+    if (jsonNames == null) {
+      return protoMsg.build();
+    }
     int matchedFields = 0;
-    // JsonNames will be null if the JSONObject is empty.
-    if (jsonNames != null) {
-      for (int i = 0; i < jsonNames.length; i++) {
-        String jsonName = jsonNames[i];
-        String jsonLowercaseName = jsonName.toLowerCase();
-        String currentScope = jsonScope + "." + jsonName;
-        FieldDescriptor field = protoSchema.findFieldByName(jsonLowercaseName);
-        if (field == null) {
-          if (!allowUnknownFields) {
-            throw new IllegalArgumentException(
-                String.format(
-                    "JSONObject has fields unknown to BigQuery: %s. Set allowUnknownFields to True to allow unknown fields.",
-                    currentScope));
-          } else {
-            continue;
-          }
-        }
-        matchedFields++;
-        if (!field.isRepeated()) {
-          fillField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+    for (int i = 0; i < jsonNames.length; i++) {
+      String jsonName = jsonNames[i];
+      // We want lowercase here to support case-insensitive data writes.
+      // The protobuf descriptor that is used is assumed to have all lowercased fields
+      String jsonLowercaseName = jsonName.toLowerCase();
+      String currentScope = jsonScope + "." + jsonName;
+      FieldDescriptor field = protoSchema.findFieldByName(jsonLowercaseName);
+      if (field == null) {
+        if (!allowUnknownFields) {
+          throw new IllegalArgumentException(
+              String.format(
+                  "JSONObject has fields unknown to BigQuery: %s. Set allowUnknownFields to True to allow unknown fields.",
+                  currentScope));
         } else {
-          fillRepeatedField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+          continue;
         }
       }
+      matchedFields++;
+      if (!field.isRepeated()) {
+        fillField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+      } else {
+        fillRepeatedField(protoMsg, field, json, jsonName, currentScope, allowUnknownFields);
+      }
     }
+
     if (matchedFields == 0 && topLevel) {
       throw new IllegalArgumentException(
           "There are no matching fields found for the JSONObject and the protocol buffer descriptor.");
@@ -120,6 +126,9 @@ public class JsonToProtoMessage {
           String.format(
               "JSONObject does not have the required field %s.%s.", jsonScope, missingFieldName));
     }
+    if (topLevel && msg.getSerializedSize() == 0) {
+      throw new IllegalArgumentException("The created protobuf message is empty.");
+    }
     return msg;
   }
 
@@ -129,7 +138,7 @@ public class JsonToProtoMessage {
    * @param protoMsg The protocol buffer message being constructed
    * @param fieldDescriptor
    * @param json
-   * @param actualJsonKeyName Actual key name in JSONObject instead of lowercased version
+   * @param exactJsonKeyName Actual key name in JSONObject instead of lowercased version
    * @param currentScope Debugging purposes
    * @param allowUnknownFields Ignores unknown JSON fields.
    * @throws IllegalArgumentException when JSON data is not compatible with proto descriptor.
@@ -138,7 +147,7 @@ public class JsonToProtoMessage {
       DynamicMessage.Builder protoMsg,
       FieldDescriptor fieldDescriptor,
       JSONObject json,
-      String actualJsonKeyName,
+      String exactJsonKeyName,
       String currentScope,
       boolean allowUnknownFields)
       throws IllegalArgumentException {
@@ -146,13 +155,13 @@ public class JsonToProtoMessage {
     try {
       switch (fieldDescriptor.getType()) {
         case BOOL:
-          protoMsg.setField(fieldDescriptor, new Boolean(json.getBoolean(actualJsonKeyName)));
+          protoMsg.setField(fieldDescriptor, new Boolean(json.getBoolean(exactJsonKeyName)));
           break;
         case BYTES:
-          protoMsg.setField(fieldDescriptor, json.getString(actualJsonKeyName).getBytes());
+          protoMsg.setField(fieldDescriptor, json.getString(exactJsonKeyName).getBytes());
           break;
         case INT64:
-          val = json.get(actualJsonKeyName);
+          val = json.get(exactJsonKeyName);
           if (val instanceof Integer) {
             protoMsg.setField(fieldDescriptor, new Long((Integer) val));
           } else if (val instanceof Long) {
@@ -162,7 +171,7 @@ public class JsonToProtoMessage {
           }
           break;
         case INT32:
-          val = json.get(actualJsonKeyName);
+          val = json.get(exactJsonKeyName);
           if (val instanceof Integer) {
             protoMsg.setField(fieldDescriptor, new Integer((Integer) val));
           } else {
@@ -170,10 +179,10 @@ public class JsonToProtoMessage {
           }
           break;
         case STRING:
-          protoMsg.setField(fieldDescriptor, json.getString(actualJsonKeyName));
+          protoMsg.setField(fieldDescriptor, json.getString(exactJsonKeyName));
           break;
         case DOUBLE:
-          val = json.get(actualJsonKeyName);
+          val = json.get(exactJsonKeyName);
           if (val instanceof Double) {
             protoMsg.setField(fieldDescriptor, new Double((double) val));
           } else if (val instanceof Float) {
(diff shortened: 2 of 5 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,482 @@
...
+          JsonToProtoMessage.convertJsonToProtoMessage(Int64Type.getDescriptor(), json);
+    } catch (IllegalArgumentException e) {
+      assertEquals(
+          e.getMessage(),
+          "There are no matching fields found for the JSONObject and the protocol buffer descriptor.");
+    }
+  }
+
+  @Test
+  public void testTopLevelMatchSecondLevelNoMatch() throws Exception {
```
yirutang, line 458: Should this be a failed case?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,482 @@
...
+    json.put("double", 1.1);
+    json.put("string", "hello");
+
+    try {
+      DynamicMessage protoMsg =
+          JsonToProtoMessage.convertJsonToProtoMessage(RepeatedInt64.getDescriptor(), json);
+    } catch (IllegalArgumentException e) {
+      assertEquals(
+          e.getMessage(),
+          "JSONObject has unknown fields. Set allowUnknownFields to True to ignore unknown fields.");
```
yirutang, line 439: Let's put the error message like this: JSONObject has fields unknown to BigQuery: f1. Set allowUnknownFields to True to allow unknown fields.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,341 @@
...
+public class JsonToProtoMessage {
+
+  /**
+   * Converts Json data to protocol buffer messages given the protocol buffer descriptor.
+   *
+   * @param protoSchema
+   * @param json
+   * @throws IllegalArgumentException when JSON data is not compatible with proto descriptor.
+   */
+  public static DynamicMessage convertJsonToProtoMessage(Descriptor protoSchema, JSONObject json)
```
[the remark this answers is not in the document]
allenc3: This one is to allow users to just pass in data without worrying about allowUnknownFields, which is the same as we did with SchemaCompatibility. Should I still remove this?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,341 @@
...
+        throw new IllegalArgumentException(
+            "JSONObject has unknown fields. Set allowUnknownFields to True to ignore unknown fields.");
+      }
+    }
+
+    for (FieldDescriptor field : protoFields) {
+      String fieldName = field.getName();
+      String currentScope = jsonScope + "." + fieldName;
+
+      if (!json.has(fieldName)) {
```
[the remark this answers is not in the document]
allenc3: Got it

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: Got it

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: This one should not be a failed case since the top level has matching fields (both test_int and complexLvl2) but the second level which is the optional field test_int in ComplexLvl2 did not match. However since the top level matched already, it doesn't matter if the second level had any matching fields. The json data is legal too since the test_int in ComplexLvl2 is optional.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: Does the f1 mean anything?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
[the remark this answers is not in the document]
yirutang: Doesn't matter too much. It is just that this library is unlikely to be used by user directly. For Schema compact check, I think we did that to avoid too many code updates.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
yirutang: But is the no_match field counted as an unknown field?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,406 @@
...
+    if (JSONObject.getNames(json).length > protoFields.size()) {
+      if (!allowUnknownFields) {
+        throw new IllegalArgumentException(
+            "JSONObject has fields unknown to BigQuery: f1. Set allowUnknownFields to True to allow unknown fields.");
+      }
+    }
+    HashMap<String, String> jsonLowercaseNameToActualName = new HashMap<String, String>();
+    String[] actualNames = JSONObject.getNames(json);
+    for (int i = 0; i < actualNames.length; i++) {
+      jsonLowercaseNameToActualName.put(actualNames[i].toLowerCase(), actualNames[i]);
```
yirutang, line 87: Instead of creating a new container, maybe use case insenstive set comparison directly?
https://stackoverflow.com/questions/24558456/java-setstring-equality-ignore-case

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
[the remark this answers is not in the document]
allenc3: Ok!

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
yirutang: No it is just an example.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: So I think I still need the HashMap to hold a mapping from <LowercasedName, Name> because in order retrieve the vjson alue, I still need to use the original name since I need to call JSONObject.get(<original name>).

yirutang: Can it be:
 Set<String> jsonNames = new TreeSet<String>(String.CASE_INSENSITIVE_ORDER);
 jsonNames.addAll(JSONObject.getName(json));

...
jsonNames.contains(field.getName())

?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,406 @@
...
+          if (val instanceof Byte) {
+            protoMsg.setField(fieldDescriptor, new Long((Byte) val));
+          } else if (val instanceof Short) {
+            protoMsg.setField(fieldDescriptor, new Long((Short) val));
+          } else if (val instanceof Integer) {
+            protoMsg.setField(fieldDescriptor, new Long((Integer) val));
+          } else if (val instanceof Long) {
+            protoMsg.setField(fieldDescriptor, new Long((Long) val));
+          } else {
+            throw new IllegalArgumentException(
```
weizhengli1992: Does this block equals to the following code?:

try {
        return Long.parseLong(json.getAsString());
} catch (Exception e) {
       throw new IllegalArgumentException();
}

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: Yes, and this does check if the field names are case-insensitive equal. But after checking, I need to know the actual name of the json field in order to access the json value, so I think I need to create a mapping from <LowercasedName -> Name>

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,406 @@
...
+      boolean allowUnknownFields)
+      throws IllegalArgumentException {
+
+    switch (fieldDescriptor.getType()) {
+      case BOOL:
+        try {
+          protoMsg.setField(fieldDescriptor, new Boolean(json.getBoolean(actualJsonKeyName)));
+        } catch (JSONException e) {
+          throw new IllegalArgumentException(
+              "JSONObject does not have a boolean field at " + currentScope + ".");
```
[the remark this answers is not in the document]
allenc3: Ok!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: I'm using org.json.JSONObject which I don't think has that method. Should I be using something else?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @param protoSchema
+   * @param json
+   * @param allowUnknownFields Ignores unknown JSON fields.
+   * @throws IllegalArgumentException when JSON data is not compatible with proto descriptor.
+   */
+  public static DynamicMessage convertJsonToProtoMessage(
+      Descriptor protoSchema, JSONObject json, boolean allowUnknownFields)
+      throws IllegalArgumentException {
+    if (json.length() == 0) {
```
yirutang, line 55: null checks on entrance of the function.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+      String jsonScope,
+      boolean topLevel,
+      boolean allowUnknownFields)
+      throws IllegalArgumentException {
+
+    DynamicMessage.Builder protoMsg = DynamicMessage.newBuilder(protoSchema);
+    String[] jsonNames = JSONObject.getNames(json);
+    int matchedFields = 0;
+    // JsonNames will be null if the JSONObject is empty.
+    if (jsonNames != null) {
```
yirutang, line 84: Return on null earlier so the flow is easier to understand.
if (jsonNames == null) {
   return protoMsg.build();
}

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+      for (int i = 0; i < jsonNames.length; i++) {
+        String jsonName = jsonNames[i];
+        String jsonLowercaseName = jsonName.toLowerCase();
```
yirutang, line 87: Add a comment about why we want lower case here, i,e. column name is case insensitive.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+   * @param actualJsonKeyName Actual key name in JSONObject instead of lowercased version
+   * @param currentScope Debugging purposes
+   * @param allowUnknownFields Ignores unknown JSON fields.
+   * @throws IllegalArgumentException when JSON data is not compatible with proto descriptor.
+   */
+  private static void fillField(
+      DynamicMessage.Builder protoMsg,
+      FieldDescriptor fieldDescriptor,
+      JSONObject json,
+      String actualJsonKeyName,
```
yirutang, line 141: exactJsonKeyName?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+          } else {
+            throw new JSONException("");
+          }
+          break;
+        case INT32:
+          val = json.get(actualJsonKeyName);
+          if (val instanceof Integer) {
+            protoMsg.setField(fieldDescriptor, new Integer((Integer) val));
+          } else {
+            throw new JSONException("");
```
yirutang, line 169: We need specific error info here, this is a field type mismatch, ideally it would be expected Integer but actual is blah.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,328 @@
...
+                    "JSONObject does not have a boolean field at %s[%d].", currentScope, i));
+          }
+        }
+        break;
+      case BYTES:
+        for (int i = 0; i < jsonArray.length(); i++) {
+          try {
+            protoMsg.addRepeatedField(fieldDescriptor, jsonArray.getString(i).getBytes());
+          } catch (JSONException e) {
+            throw new IllegalArgumentException(
```
yirutang, line 250: Can these catch cases be merged?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: So I threw this empty JsonException error and caught it in the outer loop to throw the actual error which is "JSONObject does not have a <fieldtype> field at <scope>."  If I were to fill in the exception there it would be the same message "JSONObject does not have a <fieldtype> field at <scope>."

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,638 @@
...
+  @Test
+  public void testDifferentNameCasing() throws Exception {
+    JSONObject json = new JSONObject();
+    json.put("bYtE", (byte) 1);
+    json.put("SHORT", (short) 1);
+    json.put("inT", 1);
+    json.put("lONg", 1L);
+    DynamicMessage protoMsg =
+        JsonToProtoMessage.convertJsonToProtoMessage(TestInt64.getDescriptor(), json, false);
+    AreMatchingFieldsFilledIn(protoMsg, json);
```
weizhengli1992, line 237: Seems we are using a function "AreMatchingFieldsFilledIn" to verify whether proto equals to json. What if there's a bug inside AreMatchingFieldsFilledIn?.

Can we assert protoMsg to a pre-defined proto?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,322 @@
...
+    } catch (JSONException e) {
+      throw new IllegalArgumentException(
+          "JSONObject does not have a array field at " + currentScope + ".");
+    }
+    java.lang.Object val;
+    int index = -1;
+    try {
+      switch (fieldDescriptor.getType()) {
+        case BOOL:
+          for (int i = 0; i < jsonArray.length(); i++) {
```
weizhengli1992, line 246: can we move for loop outside of switch? so that we don't need these for loops?

like :
for (int i = 0; i < jsonArray.length(); i++) {
  switch (fieldDescriptor.getType()) {
  }
}

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,322 @@
...
+        case INT64:
+          for (int i = 0; i < jsonArray.length(); i++) {
+            index = i;
+            val = jsonArray.get(i);
+            if (val instanceof Integer) {
+              protoMsg.addRepeatedField(fieldDescriptor, new Long((Integer) val));
+            } else if (val instanceof Long) {
+              protoMsg.addRepeatedField(fieldDescriptor, new Long((Long) val));
+            } else {
+              throw new JSONException("");
```
weizhengli1992, line 266: is this empty error message is intended?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,322 @@
...
+            }
+          }
+          break;
+        case INT32:
+          for (int i = 0; i < jsonArray.length(); i++) {
+            index = i;
+            val = jsonArray.get(i);
+            if (val instanceof Integer) {
+              protoMsg.addRepeatedField(fieldDescriptor, new Integer((Integer) val));
```
weizhengli1992, line 275: can new Integer((Integer)  val) just be new Integer(val) or (Integer) val?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,322 @@
...
+    } catch (UninitializedMessageException e) {
+      String errorMsg = e.getMessage();
+      int idxOfColon = errorMsg.indexOf(":");
+      String missingFieldName = errorMsg.substring(idxOfColon + 2);
+      throw new IllegalArgumentException(
+          String.format(
+              "JSONObject does not have the required field %s.%s.", jsonScope, missingFieldName));
+    }
+    if (topLevel && msg.getSerializedSize() == 0) {
+      throw new IllegalArgumentException("The created protobuf message is empty.");
```
weizhengli1992, line 130: out of curiosity, when can we hit this condition? I thought we already have "matchedFields == 0 && topLevel" check above?

allenc3: So if there's only 1 repeated field, and the json passes in an empty array, then it will trigger this condition.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
```diff
@@ -0,0 +1,322 @@
...
+      if (field == null) {
+        if (!allowUnknownFields) {
+          throw new IllegalArgumentException(
+              String.format(
+                  "JSONObject has fields unknown to BigQuery: %s. Set allowUnknownFields to True to allow unknown fields.",
+                  currentScope));
+        } else {
+          continue;
+        }
+      }
```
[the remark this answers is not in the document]
allenc3: I think I still need the else {} in order to skip this iteration when field == null and allowUnknownFields.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: Yup, good catch!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessage.java
allenc3: Yes I'll change that now.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: The proto message produced does use a pre-defined proto, so it must be of the type of the pre-defined proto. For example, for whatever protoMsg produced, if we call the function getDescriptorForType(), it will be equal to the protobuf descriptor that was passed in to create it. The AreMatchingFieldsFilledIn() function tests for field existence and if the field value is the same as the json value.

allenc3: Changed testcases to compare generated proto message with self defined proto message.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,741 @@
...
+    TestInt64 correctProto =
+        TestInt64.newBuilder().setByte(1).setShort(1).setInt(1).setLong(1).build();
+    JSONObject json = new JSONObject();
+    json.put("byte", (byte) 1);
+    json.put("short", (short) 1);
+    json.put("int", 1);
+    json.put("long", 1L);
+    DynamicMessage protoMsg =
+        JsonToProtoMessage.convertJsonToProtoMessage(TestInt64.getDescriptor(), json, false);
+    assertEquals(protoMsg, correctProto);
```
weizhengli1992, line 283: [nit] correctProto -> expectedProto

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,638 @@
...
+                  (long) Byte.MAX_VALUE,
+                  (long) Byte.MIN_VALUE,
+                  0L
+                })),
+    new JSONObject()
+        .put(
+            "test_repeated",
+            new JSONArray(
+                new Integer[] {
+                  (int) Integer.MAX_VALUE,
```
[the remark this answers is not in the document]
allenc3: Done!

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: Done!

google-cloud-bigquerystorage/pom.xml
```diff
@@ -108,6 +108,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20200518</version>
```
stephaniewang526, line 114: Please manage the version in `<dependencyManagement>` section of the root pom, instead of here. This is for long-term maintenance purpose.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
```diff
@@ -0,0 +1,751 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigquery.storage.v1alpha2;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
stephaniewang526, line 19: Please only import used functions instead of `*`

google-cloud-bigquerystorage/pom.xml
allenc3: Got it.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonToProtoMessageTest.java
allenc3: Got it.

PR #436, opened 2020-07-21
fix: add missing resource name deps to v1beta1

noahdietz:
These dependencies were not added to the `pom.xml` by the generator when resource name classes were added. I'm not sure why that is, but it will fix the CI errors in #435. I tested this locally and noted it in the config change.


noahdietz:
cc @chingor13 @kmjung

kmjung:
cc @stephaniewang526

noahdietz:
I don't have merge permissions it seems, so feel free to whenever you're ready.

PR #437, opened 2020-07-21
feature: Added JsonStreamWriter, which is built on top of a StreamWriter and writes Json data by first converting to protobuf messages then calling StreamWriter's append

allenc3:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A StreamWriter that can write JSON data (JSONObjects) to BigQuery tables. The JsonStreamWriter is
+ * built on top of a StreamWriter, and it simply converts all JSON data to protobuf messages then
+ * calls StreamWriter's append() method to write to BigQuery tables. It maintains all StreamWriter
+ * functions, but also provides an additional feature: schema update support, where if the BigQuery
+ * table schema is updated, the JsonStreamWriter is guaranteed to be updated after some time.
```
yirutang, line 40: where if the BigQuery table schema is updated, users will be able to ingest data on the new schema after some time (in order of minutes).

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+      throws Descriptors.DescriptorValidationException, IllegalArgumentException, IOException,
+          InterruptedException {
+    Matcher matcher = streamPattern.matcher(builder.streamName);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException("Invalid stream name: " + builder.streamName);
+    }
+
+    this.streamName = builder.streamName;
+    this.client = builder.client;
+    this.BQTableSchema = builder.BQTableSchema;
```
yirutang, line 69: This violates Java member variable naming?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+   *
+   * @return Table.TableSchema
+   */
+  public Table.TableSchema getBQTableSchema() {
+    return this.BQTableSchema;
+  }
+
+  /** Sets all StreamWriter settings. */
+  private void setStreamWriterSettings(
+      StreamWriter.Builder builder,
```
yirutang, line 187: Mark these as nullable?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+  }
+
+  /**
+   * newBuilder that constructs a JsonStreamWriter builder with BigQuery client being initialized by
+   * StreamWriter by default.
+   *
+   * @param streamName name of the stream that must follow
+   *     "projects/[^/]+/datasets/[^/]+/tables/[^/]+/streams/[^/]+"
+   * @param BQTableSchema schema used to convert Json to proto messages.
```
yirutang, line 220: The schema of the table when the stream was created, which is passed back through {@code WriteStream}

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+   * Updates the descriptor and BQTable schema. This function is used to support schema updates, and
+   * is called whenever AppendRowsResponse includes an updated schema. The function is synchronized
+   * since it is called through a callback (which is in another thread). If the main thread is
+   * calling append while the callback thread calls refreshAppend(), this might cause some issues.
+   *
+   * @param BQTableSchema The updated table schema.
+   */
+  private synchronized void updateDescriptor(Table.TableSchema updatedSchema)
+      throws Descriptors.DescriptorValidationException, IOException, InterruptedException {
+    if (!this.BQTableSchema.equals(updatedSchema)) {
```
yirutang, line 127: I don't think this equal check will work. One thing is that you could do a content hash and then double check on the hash. Or just assume the backend always do the write thing.

Also log some message here since this is a major operation.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Changed!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Changed BQTableSchema to be tableSchema.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it, I'll just assume the backend does the right thing. I was just thinking to eliminate processing the same update schema if it gets passed twice. Will add logging!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Done!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+      this.descriptor =
+          BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
+      this.BQTableSchema = updatedSchema;
```
weizhengli1992, line 130: do we need to add a lock for BQTableSchema. Otherwise, other threads may sill use the old schema during this process?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+  public Descriptor getDescriptor() {
+    return this.descriptor;
+  }
+
+  /**
+   * Gets current BQTableSchema
+   *
+   * @return Table.TableSchema
+   */
+  public Table.TableSchema getBQTableSchema() {
```
weizhengli1992, line 181: [nit] just name it getTableSchema?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+  /**
+   * newBuilder that constructs a JsonStreamWriter builder with BigQuery client being initialized by
+   * StreamWriter by default.
+   *
+   * @param streamName name of the stream that must follow
+   *     "projects/[^/]+/datasets/[^/]+/tables/[^/]+/streams/[^/]+"
+   * @param BQTableSchema schema used to convert Json to proto messages.
+   * @return Builder
+   */
+  public static Builder newBuilder(String streamName, Table.TableSchema BQTableSchema) {
```
weizhengli1992, line 223: Maybe better to name it as JsonWriterBuilder? So reader could easily understand what this Builder referring to?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
weizhengli1992: Better to make them optional instead of nullable.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+      return this;
+    }
+
+    /**
+     * Setter for the underlying StreamWriter's Endpoint.
+     *
+     * @param endpoint
+     * @return Builder
+     */
+    public Builder setEndpoint(String endpoint) {
```
weizhengli1992, line 343: what does this endpoint referring to? The BQ api address?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,359 @@
...
+  private static String streamPatternString =
+      "projects/[^/]+/datasets/[^/]+/tables/[^/]+/streams/[^/]+";
+  private static Pattern streamPattern = Pattern.compile(streamPatternString);
+  private static final Logger LOG = Logger.getLogger(JsonStreamWriter.class.getName());
+
+  private BigQueryWriteClient client;
+  private String streamName;
+  private StreamWriter streamWriter;
+  private Descriptor descriptor;
+  private Table.TableSchema BQTableSchema;
```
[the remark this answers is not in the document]
allenc3: Done!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Done!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: It's a static method so users will call it as JsonStreamWriter.newBuilder(), and it seems to be conventional to use newBuilder() (used in StreamWriter, Table.TableSchema, or when constructing protos) so I went with this. Should I still change the name?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Think so, StreamWriter had a setEndpointer() method so I included it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Made Append() and updateDescriptor() both synchronized methods.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
weizhengli1992: NVM.. looks good to me then

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,381 @@
...
+    return appendResponseFuture;
+  }
+
+  /**
+   * Updates the descriptor and BQTable schema. This function is used to support schema updates, and
+   * is called whenever AppendRowsResponse includes an updated schema. The function is synchronized
+   * since it is called through a callback (which is in another thread). If the main thread is
+   * calling append while the callback thread calls refreshAppend(), this might cause some issues.
+   */
+  private synchronized void updateDescriptor() {
```
yirutang, line 155: Why not directly pass in a callback to stream writer for it to execute action on schema update? The code could be simpler and there is no need to add callbacks on the response future.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -204,6 +207,81 @@ public void testBatchWriteWithCommittedStream()
     assertEquals(false, iter.hasNext());
   }
 
+  @Test
+  public void testJsonStreamWriterBatchWriteWithCommittedStream()
```
yirutang, line 211: One e2e test about schema update handling?

allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -800,6 +839,18 @@ public void onResponse(AppendRowsResponse response) {
         if (response == null) {
           inflightBatch.onFailure(new IllegalStateException("Response is null"));
         }
+        if (response.hasUpdatedSchema()) {
+          if (streamWriter.getOnSchemaUpdateRunnable() != null) {
+            streamWriter.getOnSchemaUpdateRunnable().setUpdatedSchema(response.getUpdatedSchema());
+            try {
+              streamWriter.getOnSchemaUpdateRunnable().getStreamWriter().refreshAppend();
```
yirutang, line 846: Should it just be called on the current streamwriter?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -183,6 +192,27 @@ public String getTableNameString() {
     return tableName;
   }
 
+  /* Once updated schema is retrieved, it will be set to null (matches backend implementation of clearing updated schema once it puts it in an AppendRowsResponse)
+   *
+   * @return Table.TableSchema
+   */
+  public Table.TableSchema getUpdatedSchema() {
```
yirutang, line 199: Do we need this function?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -183,6 +192,27 @@ public String getTableNameString() {
...
+    return this.updatedSchema;
+  }
+
+  /* Setter for updated schema.
+   *
+   * @param updatedSchema
+   */
+  public void setUpdatedSchema(Table.TableSchema updatedSchema) {
```
yirutang, line 207: Thread safe?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
yirutang: The execution of the Runnable and refresh append could race. I think we should simply run the runnable and let user handle everything.

allenc3: Sure, that would solve it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
allenc3: I'll make it synchronized.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
allenc3: This was to pass out the schema to the JsonStreamWriter, and technically it's not needed, just thought of letting the users see the schema. I'll remove them then.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+      builder.setOnSchemaUpdateRunnable(onSchemaUpdateRunnable);
+    }
+  }
+
+  /**
+   * Setter for descriptor.
+   *
+   * @param tableSchema
+   */
+  public synchronized void setDescriptor(Table.TableSchema tableSchema)
```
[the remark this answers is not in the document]
allenc3: Yes, this is for the runnable that is passed into StreamWriter to be able to set the descriptor for the JsonStreamWriter.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+  public Descriptor getDescriptor() {
+    return this.descriptor;
+  }
+
+  /**
+   * Gets current tableSchema
+   *
+   * @return Table.TableSchema
+   */
+  public Table.TableSchema getTableSchema() {
```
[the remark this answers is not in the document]
allenc3: Nope, will remove.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  /**
+   * Writes a JSONArray that contains JSONObjects to the BigQuery table by first converting the JSON
+   * data to protobuf messages, then using StreamWriter's append() to write the data. This will
+   * return a ApiFuture<AppendRowsResponse>, which will be processed by a callback to determine if a
+   * schema update is required.
+   *
+   * @param jsonArr The JSON array that contains JSONObjects to be written
+   * @param offset Offset for deduplication
```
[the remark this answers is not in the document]
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
```diff
@@ -0,0 +1,97 @@
...
+ * "Schema update fail: updated schema could not be converted to a valid descriptor.");
+ * return;
+ * }
+ *
+ * LOG.info("Successfully updated schema: " + this.updatedSchema);
+ * }
+ * </code>
+ * </pre>
+ */
+public abstract class OnSchemaUpdateRunnable implements Runnable {
```
weizhengli1992, line 50: This will throw NPE if Run() is called before the all necessary variable is set.

I would recommend you to have a RunInternal() function and let user to override RunInternal() instead.

So it would like:

Run() {
    Preconditions.checkNotNull(...);
    RunInternal();
}

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
```diff
@@ -0,0 +1,97 @@
...
+    this.streamWriter = streamWriter;
+  }
+
+  /**
+   * Setter for the jsonStreamWriter
+   *
+   * @param jsonStreamWriter
+   */
+  public void setJsonStreamWriter(JsonStreamWriter jsonStreamWriter) {
+    this.jsonStreamWriter = jsonStreamWriter;
```
weizhengli1992, line 80: Preconditions.checkNotNull?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,369 @@
...
+   * @param streamName name of the stream that must follow
+   *     "projects/[^/]+/datasets/[^/]+/tables/[^/]+/streams/[^/]+"
+   * @param tableSchema The schema of the table when the stream was created, which is passed back
+   *     through {@code WriteStream}
+   * @param client
+   * @return Builder
+   */
+  public static Builder newBuilder(
+      String streamName, Table.TableSchema tableSchema, BigQueryWriteClient client) {
+    Preconditions.checkNotNull(streamName, "StreamName is null.");
```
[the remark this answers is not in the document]
allenc3: I don't see a checkNotEmpty function for Preconditions. For handling the empty case though, I do check if the StreamName is properly formatted later on in the JsonStreamWriter constructor.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
allenc3: So there are 3 variables: JsonStreamWriter, StreamWriter, and the UpdatedSchema. The JsonStreamWriter and StreamWriter are both set when the JsonStreamWriter is constructed, and the UpdatedSchema is set before the Runnable is scheduled. So I think the execution order should make is so those variables will be null. What do you think?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
allenc3: So this is actually for the JsonStreamWriter to set itself when it takes in the Runnable, and not for the users to do it. Is there a way to make it so that JsonStreamWriter can do it and not the users? Would making it package-level solve this?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -800,6 +837,13 @@ public void onResponse(AppendRowsResponse response) {
         if (response == null) {
           inflightBatch.onFailure(new IllegalStateException("Response is null"));
         }
+        if (response.hasUpdatedSchema()) {
+          if (streamWriter.getOnSchemaUpdateRunnable() != null) {
```
weizhengli1992, line 841: Is it allowed to have OnSchemaUpdateRunnable here? Or we should throw exception?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
weizhengli1992: gotcha... sounds good.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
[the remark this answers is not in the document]
weizhengli1992: sounds good.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,384 @@
...
+      return this;
+    }
+
+    /**
+     * Setter for the underlying StreamWriter's CredentialsProvider.
+     *
+     * @param credentialsProvider
+     * @return Builder
+     */
+    public Builder setCredentialsProvider(CredentialsProvider credentialsProvider) {
```
weizhengli1992, line 307: are these necessary (CredentialsProvider, BatchingSettings, setRetrySettings....)? If they are, should we move them to constructor to avoid the case that user forgot to set them?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
allenc3: Users can not set onSchemaUpdateRunnable and basically not handle what to do for schema updates.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Yes these settings are necessary cause they can customize the StreamWriter's function. And so here I'm trying to let users set the settings in the JsonStreamWriter Builder class instead of the JsonStreamWriter class because it seems to make more sense having the Builder class to handle all settings. If the user's don't set them, there won't be any exceptions and the StreamWriter will be created by default settings.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
```diff
@@ -0,0 +1,601 @@
...
+        expectedProto.toByteString());
+    writer.close();
+  }
+
+  private final OnSchemaUpdateRunnable ON_SCHEMA_UPDATE_RUNNABLE =
+      new OnSchemaUpdateRunnable() {
+        public void run() {
+          this.getStreamWriter().setUpdatedSchema(this.getUpdatedSchema());
+          try {
+            this.getJsonStreamWriter().setDescriptor(this.getUpdatedSchema());
```
yirutang, line 389: Move this definition in test to the JsonWriter. Make sure the connection will always be consistent with the json schema.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * LOG.info("Successfully updated schema: " + this.getUpdatedSchema());
+ * }
+ * </code>
+ * </pre>
+ */
+public abstract class OnSchemaUpdateRunnable implements Runnable {
+  private JsonStreamWriter jsonStreamWriter;
+  private StreamWriter streamWriter;
+  private Table.TableSchema updatedSchema;
```
yirutang, line 58: Add a comment that this is used for test purpose.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -110,6 +110,10 @@
 
   private Integer currentRetries = 0;
 
+  // Used for schema updates
+  private Table.TableSchema updatedSchema;
```
yirutang, line 114: guard this by a lock.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,384 @@
...
+
+    /**
+     * Setter for the action to perform when there is a schema update.
+     *
+     * @param onSchemaUpdateRunnable An abstract class that implements runnable and provides access
+     *     to the JsonStreamWriter, the StreamWriter, and the updated schema. Users should implement
+     *     the run() function.
+     * @return Builder
+     */
+    public Builder setOnSchemaUpdateRunnable(OnSchemaUpdateRunnable onSchemaUpdateRunnable) {
```
yirutang, line 367: You should define your SchemaRunnable inside JsonWriter instead of letting user to define it. User shouldn't be worried about SchemaUpdate at all.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
```diff
@@ -0,0 +1,601 @@
+/*
```
yirutang, line 1: Add a test case that multiple threads calling JsonWriter to append at the same time.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
allenc3: The updatedSchema in the OnSchemaUpdateRunnable is used so users can customize what they are doing with it in the runnable. The table schema in the StreamWriter and jsonStreamWriter is for test purposes only.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
allenc3: Also decided to remove TableSchema entirely in StreamWriter and JsonStreamWriter, and just let JsonStreamWriter hold on to a descriptor.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
allenc3: Removed this.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
allenc3: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,377 @@
...
+    private RetrySettings retrySettings;
+    private ExecutorProvider executorProvider;
+    private String endpoint;
+    private OnSchemaUpdateRunnable onSchemaUpdateRunnable;
+
+    private final OnSchemaUpdateRunnable ON_SCHEMA_UPDATE_RUNNABLE =
+        new OnSchemaUpdateRunnable() {
+          public void run() {
+            try {
+              this.getStreamWriter().refreshAppend();
```
yirutang, line 267: You need to make refreshAppend and setDescriptor go together, preventing any append actions in between. Because if there is any Append in between, then the setDescriptor is too late.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
```diff
@@ -0,0 +1,611 @@
...
+
+    // Check if writer schemas were added in for both connections.
+    assertTrue(testBigQueryWrite.getAppendRequests().get(0).getProtoRows().hasWriterSchema());
+    assertTrue(testBigQueryWrite.getAppendRequests().get(1).getProtoRows().hasWriterSchema());
+
+    writer.close();
+  }
+
+  @Test
+  public void testMultiThreadAppend() throws Exception {
```
yirutang, line 545: How about drop in some schema update in between these parallel calls.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,377 @@
...
+    // IllegalArgumentException/IllegalStateException/NullPointerException and will halt processing
+    // of JSON data.
+    for (int i = 0; i < jsonArr.length(); i++) {
+      JSONObject json = jsonArr.getJSONObject(i);
+      Message protoMessage =
+          JsonToProtoMessage.convertJsonToProtoMessage(this.descriptor, json, allowUnknownFields);
+      rowsBuilder.addSerializedRows(protoMessage.toByteString());
+    }
+    AppendRowsRequest.ProtoData.Builder data = AppendRowsRequest.ProtoData.newBuilder();
+    data.setWriterSchema(ProtoSchemaConverter.convert(this.descriptor));
```
yirutang, line 134: Don't append writer schema to the request for every request. It only needs to be set for the first request. You can see how DirectWriter handles this.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,377 @@
...
+   * data to protobuf messages, then using StreamWriter's append() to write the data. If there is a
+   * schema update, the OnSchemaUpdateRunnable will be used to determine what actions to perform.
+   *
+   * @param jsonArr The JSON array that contains JSONObjects to be written
+   * @param offset Offset for deduplication
+   * @param allowUnknownFields if true, json data can have fields unknown to the BigQuery table.
+   * @return ApiFuture<AppendRowsResponse> returns an AppendRowsResponse message wrapped in an
+   *     ApiFuture
+   */
+  public synchronized ApiFuture<AppendRowsResponse> append(
```
yirutang, line 121: Is there a reason for it to be a synchronized method? Multiple append can happen at the same time right?

When running together with schema update, maybe only a reader lock is needed for append and a writer lock for schema update method, so that no append will go between refresh and set update schema.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
weizhengli1992: Do we also need to set AppendRowsRequest.write_stream?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+    AppendRowsRequest.ProtoData.Builder data = AppendRowsRequest.ProtoData.newBuilder();
+    data.setWriterSchema(ProtoSchemaConverter.convert(this.descriptor));
+    data.setRows(rowsBuilder.build());
+
+    final ApiFuture<AppendRowsResponse> appendResponseFuture =
+        this.streamWriter.append(
+            AppendRowsRequest.newBuilder()
+                .setProtoRows(data.build())
+                .setOffset(Int64Value.of(offset))
```
weizhengli1992, line 141: Do we also need to set AppendRowsRequest.write_stream?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+package com.google.cloud.bigquery.storage.v1alpha2;
+
+/**
+ * A abstract class that implements the Runable interface and provides access to the current
+ * JsonStreamWriter, StreamWriter, and updatedSchema. This runnable will only be called when a
+ * updated schema has been passed back through the AppendRowsResponse. Users should only implement
+ * the run() function. The following example performs a simple schema update.
+ *
+ * <p>Performing a schema update requires 2 steps: making a new connection with the same WriteStream
```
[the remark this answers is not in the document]
allenc3: I was thinking it would be hard for users to come up with what methods to use for schema updates unless they know in detail what the code is, so just wanted to put a case of schema update on there so users can build their own schema update based on the simple one. Should I still remove it?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -288,6 +301,12 @@ public void refreshAppend() throws IOException, InterruptedException {
       }
     } catch (InterruptedException expected) {
     }
+    // Currently there is a bug that it took reconnected stream 5 seconds to pick up
+    // stream count. So wait at least 7 seconds before sending a new request.
+    Thread.sleep(
+        Math.max(
```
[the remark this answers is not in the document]
allenc3: I think this also means max. If initialRetryDelay is greater than 7 seconds, let it be the initialRetryDelay, else if initialRetryDelay <= 7 seconds, let it be 7 seconds.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Good catch. I'll change this.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriterTest.java
allenc3: Will do!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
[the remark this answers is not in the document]
yirutang: yeah, you are right.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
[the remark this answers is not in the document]
yirutang: Users will usually have an idea of how this should be done, they will be developers for large connectors. You only need to make it clear why you implement it this way for JsonWriter.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Resolved: StreamWriter will remove the schemas.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: StreamWriter's getMergedRequest method will call setWriteStream. Also the tests run fine (even the e2e) so I think it is already handled by the getMergedRequest method.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/OnSchemaUpdateRunnable.java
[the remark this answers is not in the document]
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Removed synchronized from method signature. Let StreamWriter handle appending the first writer schema after a refreshAppend() so the first schema sent is always right. This removes the need of refreshAppend() having to deal with calling setUpdateSchema(), which is a function specific to JsonStreamWriter and not StreamWriter.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
```diff
@@ -268,26 +296,40 @@ public void flush(long offset) {
...
+            this.retrySettings.getInitialRetryDelay().toMillis(),
+            Duration.ofSeconds(REFRESH_STREAM_WAIT_TIME).toMillis()));
+    if (this.onSchemaUpdateRunnable != null) {
```
yirutang, line 325: With the lock, do we still need this? Since SchemaUpdateRunnable is public interface, I'd like the syntax to be as simple as possible. Having an additional thing about update schema seems to be a little complicated.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    /** Getter for the jsonStreamWriter */
+    public JsonStreamWriter getJsonStreamWriter() {
+      return this.jsonStreamWriter;
+    }
+
+    @Override
+    public void run() {
+      this.setAttachUpdatedTableSchema(true);
```
yirutang, line 265: Instead of calling refreshAppend on streamwriter, could we have a method on JsonWriter that does refreshAppend and setDescriptor, so that we can lock the two operations together (with a lock that will be called by Append as well).

yirutang: We will still need locking inside StreamWriter since it is different layer of interface.

allenc3: Fixed, added a method called refreshConnection that calls refreshAppend, writeAllOutstanding, and setDescriptor.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/StreamWriter.java
allenc3: Removed and created new method refreshConnection

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,404 @@
...
+    return new Builder(streamName, tableSchema, client);
+  }
+
+  /** Closes the underlying StreamWriter. */
+  @Override
+  public void close() {
+    this.streamWriter.close();
+  }
+
+  public class JsonStreamWriterOnSchemaUpdateRunnable extends OnSchemaUpdateRunnable {
```
yirutang, line 264: This can be defined as inner class to JsonStreamWriter. It is implementation detail.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,404 @@
...
+  }
+
+  /**
+   * Refreshes connection for a JsonStreamWriter by first flushing all remaining rows, then calling
+   * refreshAppend(), and finally setting the descriptor. All of these actions need to be performed
+   * atomically to avoid having synchronization issues with append(). Flushing all rows first is
+   * necessary since if there are rows remaining when the connection refreshes, it will send out the
+   * old writer schema instead of the new one.
+   */
+  public void refreshConnection()
```
yirutang, line 154: Make is non public too.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,404 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigquery.storage.v1alpha2;
+
+import com.google.api.core.*;
```
stephaniewang526, line 18: please do not import * -- only import things we need

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,404 @@
...
+import com.google.api.gax.batching.BatchingSettings;
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.api.gax.core.ExecutorProvider;
+import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.*;
```
stephaniewang526, line 23: same as above -- please don't import *

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
```diff
@@ -0,0 +1,404 @@
...
+import com.google.cloud.bigquery.storage.v1alpha2.ProtoBufProto.ProtoRows;
+import com.google.cloud.bigquery.storage.v1alpha2.Storage.AppendRowsRequest;
+import com.google.cloud.bigquery.storage.v1alpha2.Storage.AppendRowsResponse;
+import com.google.common.base.Preconditions;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Int64Value;
+import com.google.protobuf.Message;
+import java.io.IOException;
+import java.util.*;
```
stephaniewang526, line 33: same as above -- please don't import *

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: Got it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/JsonStreamWriter.java
allenc3: I went ahead and removed all the import * from other files as well.

PR #522, opened 2020-09-04
fix:update schema compat check with backend type changes

yirutang:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1alpha2/SchemaCompatibility.java
```diff
@@ -348,6 +343,9 @@ private void protoFieldTypeIsCompatibleWithBQFieldType(
       case "NUMERIC":
         match = isCompatibleWithBQNumeric(protoType);
         break;
+      case "BIGNUMERIC":
+        match = isCompatibleWithBQNumeric(protoType);
```
stephaniewang526, line 347: please add test for this

yirutang: Removed it for now, it seems the value is not available in external library yet.

stephaniewang526 (approved):

PR #551, opened 2020-09-22
deps: update arrow-vector to v1

stephaniewang526:

### DIFF
--- a/samples/install-without-bom/pom.xml
+++ b/samples/install-without-bom/pom.xml
@@ -42,7 +42,7 @@
     <dependency>
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-vector</artifactId>
-      <version>0.17.1</version>
+      <version>1.0.1</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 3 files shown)


samples/install-without-bom/pom.xml
```diff
@@ -42,7 +42,12 @@
     <dependency>
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-vector</artifactId>
-      <version>0.17.1</version>
+      <version>1.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.arrow</groupId>
+      <artifactId>arrow-memory-netty</artifactId>
+      <version>1.0.1</version>
```
chingor13, line 50: Do you want to use a property for this? Otherwise, you will get separate renovate PRs for each of these deps

stephaniewang526: sure - done

chingor13 (approved):
You probably want this as `samples(deps): update arrow-vector to v1`. Otherwise, an update will tell release-please to bump the client version

PR #748, opened 2020-12-18
fix: enable tests that are disabled due to breaking change and stop ignoring ALREADY_EXISTED error

yirutang:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -29,15 +29,11 @@
 import com.google.api.gax.core.InstantiatingExecutorProvider;
 import com.google.api.gax.grpc.GrpcStatusCode;
 import com.google.api.gax.retrying.RetrySettings;
-import com.google.api.gax.rpc.AbortedException;
-import com.google.api.gax.rpc.BidiStreamingCallable;
-import com.google.api.gax.rpc.ClientStream;
-import com.google.api.gax.rpc.ResponseObserver;
-import com.google.api.gax.rpc.StreamController;
-import com.google.api.gax.rpc.TransportChannelProvider;
+import com.google.api.gax.rpc.*;
```
stephaniewang526, line 32: We should just import the packages we need instead of using `*` please.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -485,17 +481,16 @@ private void onFailure(Throwable t) {
...
-        //   }
-        //  singleResponse.setOffset(actualOffset);
-        // }
+        if (response.getAppendResult().hasOffset()) {
+          long actualOffset = response.getAppendResult().getOffset().getValue();
+          for (int j = 0; j < i; j++) {
+            actualOffset +=
+                inflightRequests.get(j).message.getProtoRows().getRows().getSerializedRowsCount();
+          }
+          LOG.info("Acutal offset:" + actualOffset);
```
stephaniewang526, line 490: Just a note that eventually we want to make these logger output `FINER` or `FINEST` since we don't want to overwhelm developers with granular information like this unless it is very important for them to know.

stephaniewang526 (changes requested):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
yirutang: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
yirutang: Removed.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
stephaniewang526: Looks like it is not updated: https://github.com/googleapis/java-bigquerystorage/pull/748/files#diff-683e090fb3d931a66585adce0c23b4266a62b9f47ff80e497a6ebf1e8a1ebe54R32

stephaniewang526 (approved):

PR #755, opened 2020-12-30
feat: Remove IgnoreUnknownFields support on JsonStreamWriter

yirutang:
IgnoreUnknownFields is no longer supported on API level. So make JsonStreamWriter consistent with it.


stephaniewang526:
Resolving binary compat issue in https://github.com/googleapis/java-bigquerystorage/pull/757

PR #799, opened 2021-01-14
fix: StreamWriter hang when we reach the inflight limit control and is doing a retry

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
@@ -922,41 +922,43 @@ public class StreamWriter implements AutoCloseable {
         }
         inflightBatch = this.inflightBatches.poll();
       }
-      try {
-        if (isRecoverableError(t)) {
-          try {
-            if (streamWriter.currentRetries < streamWriter.getRetrySettings().getMaxAttempts()
-                && !streamWriter.shutdown.get()) {
-              streamWriter.refreshAppend();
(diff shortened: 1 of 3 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -922,41 +922,43 @@ public void onError(Throwable t) {
...
               streamWriter.currentRetries = 0;
             }
           }
-        } else {
-          inflightBatch.onFailure(t);
-          abortInflightRequests(t);
+        } catch (IOException | InterruptedException e) {
+          LOG.info("Got exception while retrying: " + e.toString());
+          inflightBatch.onFailure(e);
+          abortInflightRequests(e);
```
stephaniewang526, line 951: can we add test coverage for catching these lines?

stephaniewang526 (changes requested):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
yirutang: good point. I tried this morning but couldn't figure out a good way to hit the code. I will play around a little more, but could we proceed with this first. Datalfow is waiting for a fix.

stephaniewang526: SG

stephaniewang526 (approved):

PR #832, opened 2021-02-16
ci: test out new sonatype google endpoint

stephaniewang526:


pom.xml
```diff
@@ -44,11 +44,11 @@
   <distributionManagement>
     <snapshotRepository>
       <id>sonatype-nexus-snapshots</id>
-      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+      <url>https://google.oss.sonatype.org/content/repositories/snapshots</url>
```
chingor13, line 47: Did we want the snapshots to also go to the new endpoint?

chingor13 (approved):

pom.xml
stephaniewang526: Yes - Joel recommended that we stay consistent in the last email. I will give this a try also.

PR #837, opened 2021-02-17
fix: Remove the appendAndRefreshLock, replace it with a less board synchronize on clientStream

yirutang:


yirutang:
I would like to block on yayi's review. Thanks!

yayi-google (commented):
This is making the race worse. The right direction is to use fewer locks, more coarse locks.

Here you are using finer locks, which is even harder to get right.

PR #849, opened 2021-02-17
fix: remove reconnection feature from client library

yirutang:
Since there is major refactoring needed in order for this to work smoothly, temporarily remove this feature and let client retry connection. Also remove a flaky test FlushAllFailed, that feature is going to be removed in another PR.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
@@ -65,8 +65,7 @@ import org.threeten.bp.Duration;
  * without offset, please use a simpler writer {@code DirectWriter}.
  *
  * <p>A {@link StreamWrier} provides built-in capabilities to: handle batching of messages;
- * controlling memory utilization (through flow control); automatic connection re-establishment and
- * request cleanup (only keeps write schema on first request in the stream).
+ * controlling memory utilization (through flow control); request cleanup (only keeps write schema on first request in the stream).
  *
  * <p>With customizable options that control:
  *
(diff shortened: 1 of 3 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -496,6 +495,7 @@ private void onFailure(Throwable t) {
         return;
       } else {
         LOG.info("Setting " + t.toString() + " on response");
+        streamWriter.messagesWaiter.release(this.getByteSize());
```
yayi-google, line 499: Why the move? You would double release resources onResponse.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -943,44 +943,8 @@ public void onError(Throwable t) {
         }
         inflightBatch = this.inflightBatches.poll();
       }
```
yayi-google, line 945: I don't think you need these code any more. You can simply abortInflightRequests(t);

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriter.java
```diff
@@ -863,14 +863,20 @@ public void onStart(StreamController controller) {
...
           InflightBatch inflightBatch = this.inflightBatches.poll();
-          inflightBatch.onFailure(
-              new AbortedException(
-                  "Request aborted due to previous failures",
-                  t,
-                  GrpcStatusCode.of(Status.Code.ABORTED),
-                  true));
+          if (first_error) {
+            inflightBatch.onFailure(t);
+            first_error = true;
```
yayi-google, line 871: first_error = false.

This means you are missing a unit test.

stephaniewang526 (approved):

PR #867, opened 2021-02-21
feat: Create Stream writer v2 - starting with core logics

yayi-google:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -23,7 +23,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Uninterruptibles;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
-import java.time.Duration;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.concurrent.TimeUnit;
@@ -52,8 +51,6 @@ import javax.annotation.concurrent.GuardedBy;
  */
 public class StreamWriterV2 implements AutoCloseable {
   private static final Logger log = Logger.getLogger(StreamWriterV2.class.getName());
-
-  private static final Duration DONE_CALLBACK_WAIT_TIMEOUT = Duration.ofMinutes(10);
 
   private Lock lock;
   private Condition hasMessageInWaitingQueue;
@@ -104,23 +101,28 @@ public class StreamWriterV2 implements AutoCloseable {
     this.waitingRequestQueue = new LinkedList<AppendRequestAndResponse>();
     this.inflightRequestQueue = new LinkedList<AppendRequestAndResponse>();
     this.streamConnection =
-        new StreamConnection(builder.client, new RequestCallback() {
-          @Override
-          public void run(AppendRowsResponse response) {
-            requestCallback(response);
-          }
-        }, new DoneCallback() {
-          @Override
-          public void run(Throwable finalStatus) {
-            doneCallback(finalStatus);
-          }
-        });
-    this.appendThread = new Thread(new Runnable() {
-      @Override
-      public void run() {
-        appendLoop();
-      }
-    });
+        new StreamConnection(
+            builder.client,
+            new RequestCallback() {
+              @Override
+              public void run(AppendRowsResponse response) {
+                requestCallback(response);
+              }
+            },
+            new DoneCallback() {
+              @Override
+              public void run(Throwable finalStatus) {
+                doneCallback(finalStatus);
+              }
+            });
+    this.appendThread =
+        new Thread(
+            new Runnable() {
+              @Override
+              public void run() {
+                appendLoop();
+              }
+            });
     this.appendThread.start();
   }
 
@@ -210,10 +212,16 @@ public class StreamWriterV2 implements AutoCloseable {
       try {
         hasMessageInWaitingQueue.await(100, TimeUnit.MILLISECONDS);
         while (!this.waitingRequestQueue.isEmpty()) {
-          localQueue.addLast(this.waitingRequestQueue.pollFirst());
+          AppendRequestAndResponse requestWrapper = this.waitingRequestQueue.pollFirst();
+          this.inflightRequestQueue.addLast(requestWrapper);
+          localQueue.addLast(requestWrapper);
         }
       } catch (InterruptedException e) {
-        log.warning("Interrupted while waiting for message. Error: " + e.toString());
+        log.warning(
+            "Interrupted while waiting for message. Stream: "
+                + streamName
+                + " Error: "
+                + e.toString());
       } finally {
         this.lock.unlock();
       }
@@ -223,16 +231,8 @@ public class StreamWriterV2 implements AutoCloseable {
       }
 
       // TODO: Add reconnection here.
-
-      this.lock.lock();
-      try {
-        while (!localQueue.isEmpty()) {
-          AppendRequestAndResponse requestWrapper = localQueue.pollFirst();
-          this.inflightRequestQueue.addLast(requestWrapper);
-          this.streamConnection.send(requestWrapper.message);
-        }
-      } finally {
-        this.lock.unlock();
+      while (!localQueue.isEmpty()) {
+        this.streamConnection.send(localQueue.pollFirst().message);
       }
     }
 
@@ -240,26 +240,7 @@ public class StreamWriterV2 implements AutoCloseable {
     // At this point, the waiting queue is drained, so no more requests.
     // We can close the stream connection and handle the remaining inflight requests.
     this.streamConnection.close();
-
-    log.info("Waiting for done callback from stream connection. Stream: " + streamName);
-    long waitDeadlineMs = System.currentTimeMillis() + DONE_CALLBACK_WAIT_TIMEOUT.toMillis();
-    while (true) {
-      if (System.currentTimeMillis() > waitDeadlineMs) {
-        log.warning(
-            "Timeout waiting for done wallback. Skip inflight cleanup. Stream: " + streamName);
-        return;
-      }
-      this.lock.lock();
-      try {
-        if (connectionFinalStatus != null) {
-          // Done callback is received, break.
-          break;
-        }
-      } finally {
-        this.lock.unlock();
-      }
-      Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
-    }
+    waitForDoneCallback();
 
     // At this point, there cannot be more callback. It is safe to clean up all inflight requests.
     log.info(
@@ -281,6 +262,22 @@ public class StreamWriterV2 implements AutoCloseable {
           && this.waitingRequestQueue.isEmpty();
     } finally {
       this.lock.unlock();
+    }
+  }
+
+  private void waitForDoneCallback() {
+    log.info("Waiting for done callback from stream connection. Stream: " + streamName);
+    while (true) {
+      this.lock.lock();
+      try {
+        if (connectionFinalStatus != null) {
+          // Done callback is received, return.
+          return;
+        }
+      } finally {
+        this.lock.unlock();
+      }
+      Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
     }
   }
 
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -1,0 +1,352 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


yayi-google:
It seems like I cannot use any nice Java8 features, like method references or lambda.

Is it because we still need to support Java7?

Do I have to write old Java7 style code instead, or is there a workaround?

Thank you!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,352 @@
...
+
+    log.info("Cleanup starts. Stream: " + streamName);
+    // At this point, the waiting queue is drained, so no more requests.
+    // We can close the stream connection and handle the remaining inflight requests.
+    this.streamConnection.close();
+
+    log.info("Waiting for done callback from stream connection. Stream: " + streamName);
+    long waitDeadlineMs = System.currentTimeMillis() + DONE_CALLBACK_WAIT_TIMEOUT.toMillis();
+    while (true) {
+      if (System.currentTimeMillis() > waitDeadlineMs) {
```
bigang-li, line 230: we can't have a real deadline here, we must wait for the done callback, otherwise the done callback could be called at any time (and may cause exception).

To make the code more readable, consider a small member function such as WaitForStreamConnectionFinished or similar around the done callback, so here we just call that function.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2Test.java
```diff
@@ -0,0 +1,283 @@
+/*
```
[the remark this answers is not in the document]
yayi-google: Yes, I already copied the relevant tests. However, most of the existing unit tests are related to flow control and batching, so they don't apply.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,352 @@
...
+  /** A builder of {@link StreamWriterV2}s. */
+  public static final class Builder {
+
+    private String streamName;
+
+    private BigQueryWriteClient client;
+
+    private Builder(String streamName, BigQueryWriteClient client) {
+      this.streamName = Preconditions.checkNotNull(streamName);
+      ;
```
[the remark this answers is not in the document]
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,352 @@
...
+        this.lock.unlock();
+      }
+
+      if (localQueue.isEmpty()) {
+        continue;
+      }
+
+      // TODO: Add reconnection here.
+
+      this.lock.lock();
```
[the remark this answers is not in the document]
yayi-google: Done.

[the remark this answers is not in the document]
yayi-google: Removed.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,352 @@
...
+    Deque<AppendRequestAndResponse> localQueue = new LinkedList<AppendRequestAndResponse>();
+    while (!waitingQueueDrained()) {
+      this.lock.lock();
+      try {
+        hasMessageInWaitingQueue.await(100, TimeUnit.MILLISECONDS);
+        while (!this.waitingRequestQueue.isEmpty()) {
+          localQueue.addLast(this.waitingRequestQueue.pollFirst());
+        }
+      } catch (InterruptedException e) {
+        log.warning("Interrupted while waiting for message. Error: " + e.toString());
```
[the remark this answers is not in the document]
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,367 @@
...
+ *
+ * <p>TODO: Add max size check.
+ *
+ * <p>TODO: Add inflight control.
+ *
+ * <p>TODO: Attach traceId.
+ *
+ * <p>TODO: Support batching.
+ *
+ * <p>TODO: Support schema change.
```
yirutang, line 50: There is another popular request, which is, the user just call append without looking at the future, and there is a FlushAll to flush the inflight queue and throw if there is any error happened. But the priority of it is behind support schema change.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,364 @@
...
+   *
+   * @param message the message in serialized format to write to BigQuery.
+   * @return the message ID wrapped in a future.
+   */
+  public ApiFuture<AppendRowsResponse> append(AppendRowsRequest message) {
+    AppendRequestAndResponse requestWrapper = new AppendRequestAndResponse(message);
+    this.lock.lock();
+    try {
+      if (userClosed) {
+        requestWrapper.appendResult.setException(
```
yirutang, line 160: Do you want to just throw directly here?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,364 @@
...
+            new StatusRuntimeException(
+                Status.fromCode(Status.Code.FAILED_PRECONDITION)
+                    .withDescription("Stream is already closed")));
+        return requestWrapper.appendResult;
+      }
+      if (connectionFinalStatus != null) {
```
yirutang, line 166: Same here.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2Test.java
```diff
@@ -0,0 +1,293 @@
...
+        TimeoutException.class,
+        new ThrowingRunnable() {
+          @Override
+          public void run() throws Throwable {
+            appendFuture1.get(1, TimeUnit.SECONDS);
+          }
+        });
+
+    // Within 2 seconds, the request should be done and stream should be closed.
+    closeThread.join(2000);
```
yirutang, line 265: I think these ways can be replaced by fake clock. The test code on the old streamwriter should work.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/FakeBigQueryWriteImpl.java
```diff
@@ -121,10 +123,16 @@ public void onNext(AppendRowsRequest value) {
...
+            if (responseSleep.compareTo(Duration.ZERO) > 0) {
+              LOG.info("Sleeping before response for " + responseSleep.toString());
+              Uninterruptibles.sleepUninterruptibly(
+                  responseSleep.toMillis(), TimeUnit.MILLISECONDS);
+            }
             if (responseDelay == Duration.ZERO) {
               sendResponse(response, responseObserver);
             } else {
               final Response responseToSend = response;
+              // TODO(yirutang): This is very wrong because it messes up response/complete ordering.
```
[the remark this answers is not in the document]
yayi-google: 1. The delay is universal, so you cannot guarantee the ordering. So response 2 can be sent before response 1.
2. When user closes the connect, server closes the connection immediately without waiting for those two callback. And callback will fail.

We can discuss it offline if you want to understand more about this.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: I don't think we want to support this at all. It seems very weird. We can talk about this offline.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: I don't mind either way, but dataflow prefers us to set the future. So I prefer to keep it this way.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: I don't mind either way, but dataflow prefers us to set the future. So I prefer to keep it this way.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2Test.java
yayi-google: Yeah, I would like to do that as well. But I cannot find a fake clock implementation in Java, that supports sleep.

yirutang: The FakeBigQueryImpl already supports the fakeclock and adding these changes would work:
https://screenshot.googleplex.com/4hPUAWG2HMctcE7

yirutang: https://screenshot.googleplex.com/6wcobhWsBJggWvV

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,364 @@
...
+
+  private void requestCallback(AppendRowsResponse response) {
+    AppendRequestAndResponse requestWrapper;
+    this.lock.lock();
+    try {
+      requestWrapper = this.inflightRequestQueue.pollFirst();
+    } finally {
+      this.lock.unlock();
+    }
+    requestWrapper.appendResult.set(response);
```
yirutang, line 314: For in stream error, should we also fail out as of now? It can be OUT_OF_RANGE error.

yayi-google: We've stated clearly in the comment of append method that user should check the in stream error. It is the same in the old stream writer as well.

I would prefer to keep this layer as thin as possible to simplify the logic. I don't too much logic in the client library, which is hard to update. It seems unnecessary to me.

However, if you believe otherwise, we could discuss offline separately, and make separate CLs. if necessary.

We don't need to block this CL because of this. Can we move forward?

yirutang: Sure, there is nothing blocking.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+  }
+
+  /** Close the stream writer. Shut down all resources. */
+  @Override
+  public void close() {
+    log.info("User closing stream: " + streamName);
+    this.lock.lock();
+    try {
+      this.userClosed = true;
```
bigang-li, line 188: adding a local var to remember the previous value, so if user closed twice, the second time will be no_op (don't call the appendThread.join() multiple times).

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -0,0 +1,364 @@
...
+                + streamName
+                + " Error: "
+                + e.toString());
+      } finally {
+        this.lock.unlock();
+      }
+
+      if (localQueue.isEmpty()) {
+        continue;
+      }
```
bigang-li, line 231: maybe be save a few lines of code here? it's useless according to context.

bigang-li (approved):

stephaniewang526 (approved):

PR #875, opened 2021-02-23
feat: Support inflight control in StreamWriterV2

yayi-google:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -22,6 +22,7 @@ import com.google.cloud.bigquery.storage.v1beta2.StreamConnection.RequestCallbac
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Uninterruptibles;
 import io.grpc.Status;
+import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
 import java.util.Deque;
 import java.util.LinkedList;
@@ -41,8 +42,6 @@ import javax.annotation.concurrent.GuardedBy;
  *
  * <p>TODO: Add max size check.
  *
- * <p>TODO: Add inflight control.
- *
  * <p>TODO: Attach traceId.
  *
  * <p>TODO: Support batching.
@@ -54,11 +53,34 @@ public class StreamWriterV2 implements AutoCloseable {
 
   private Lock lock;
   private Condition hasMessageInWaitingQueue;
+  private Condition inflightReduced;
 
   /*
    * The identifier of stream to write to.
    */
   private final String streamName;
+
+  /*
+   * Max allowed inflight requests in the stream. Method append is blocked at this.
+   */
+  private final long maxInflightRequests;
+
+  /*
+   * Max allowed inflight bytes in the stream. Method append is blocked at this.
+   */
+  private final long maxInflightBytes;
+
+  /*
+   * Tracks current inflight requests in the stream.
+   */
+  @GuardedBy("lock")
+  private long inflightRequests = 0;
+
+  /*
+   * Tracks current inflight bytes in the stream.
+   */
+  @GuardedBy("lock")
+  private long inflightBytes = 0;
 
   /*
    * Indicates whether user has called Close() or not.
@@ -97,7 +119,10 @@ public class StreamWriterV2 implements AutoCloseable {
   private StreamWriterV2(Builder builder) {
     this.lock = new ReentrantLock();
     this.hasMessageInWaitingQueue = lock.newCondition();
+    this.inflightReduced = lock.newCondition();
     this.streamName = builder.streamName;
+    this.maxInflightRequests = builder.maxInflightRequest;
+    this.maxInflightBytes = builder.maxInflightBytes;
     this.waitingRequestQueue = new LinkedList<AppendRequestAndResponse>();
     this.inflightRequestQueue = new LinkedList<AppendRequestAndResponse>();
     this.streamConnection =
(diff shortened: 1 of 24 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -336,15 +391,33 @@ private void doneCallback(Throwable finalStatus) {
...
+
+    private long maxInflightBytes = DEFAULT_MAX_INFLIGHT_BYTES;
+
     private Builder(String streamName, BigQueryWriteClient client) {
       this.streamName = Preconditions.checkNotNull(streamName);
       this.client = Preconditions.checkNotNull(client);
     }
 
+    public Builder setMaxInflightRequests(long value) {
+      this.maxInflightRequest = value;
```
yirutang, line 412: There is a formal way of presenting this:
https://screenshot.googleplex.com/8zfp4G83oeinQwC
maybe also support throw mode? I think Bigang mentioned we'd like a block with timeout but I don't see how it can be specified with FlowControlSettings. I am fine with out it for now.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2Test.java
```diff
@@ -288,6 +304,65 @@ public void serverCloseWhileRequestsInflight() throws Exception {
...
+    List<ApiFuture<AppendRowsResponse>> futures = new ArrayList<>();
+    long appendStartTimeMs = System.currentTimeMillis();
+    for (int i = 0; i < appendCount; i++) {
+      futures.add(sendTestMessage(writer, new String[] {String.valueOf(i)}));
+    }
+    long appendElapsedMs = System.currentTimeMillis() - appendStartTimeMs;
+    assertTrue(appendElapsedMs >= 1000);
+
+    for (int i = 0; i < appendCount; i++) {
+      assertEquals(i, futures.get(i).get().getAppendResult().getOffset().getValue());
```
yirutang, line 363: You code is safe, but maybe just verify the append is in order as an extra? You can assign an offset to request and get it from the server.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -171,8 +196,29 @@ public void run() {
                         "Stream is closed due to " + connectionFinalStatus.toString())));
         return requestWrapper.appendResult;
       }
+
+      ++this.inflightRequests;
+      this.inflightBytes += requestWrapper.messageSize;
       waitingRequestQueue.addLast(requestWrapper);
```
yirutang, line 202: Why do you want to check the limit after adding the request to wait queue?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: I'm not sure if it is the formal way. According to the code comments
"/** Settings for {@link FlowController}. */"

We are not using FlowController, which is some beta feature and is subject to change based on comments, so why should we use its setting object? I would prefer to use our own inflight control here.

I don't think Bigang wants to block with timeout here. It doesn't make much sense.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2Test.java
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: We could do it either way. Bigang mentioned to do it AFTER, so I just followed it to avoid conversions like this...

bigang-li: It's okay to check after adding the queue. Holding the Append() is a natural way to throttle the user when necessary.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -171,8 +196,29 @@ public void run() {
...
+              "Interrupted while waiting for inflight quota. Stream: "
+                  + streamName
+                  + " Error: "
+                  + e.toString());
+          throw new StatusRuntimeException(
+              Status.fromCode(Code.CANCELLED)
+                  .withCause(e)
+                  .withDescription("Interrupted while waiting for quota."));
+        }
+      }
```
bigang-li, line 235: Try to make the coder a little bit cleaner, let me try:

var flight_request/bytes;
{
  lock;
  addToQueue();
  inflight_requests/bytes = xxx;
}

// This should be normal cases (without extra lock operation).
if (inflight_request/bytes < threshold) {
  return;
}

// Too busy case
waitConditionWithATimeout(xxx)

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: Why is this not FlowControl? It would make the new writer's interface closer to the old interface.

yayi-google: Have you looked at the source code for FlowControlSettings?
/** Settings for {@link FlowController}. */

In FlowController.java:
/** Provides flow control capability. */
@BetaApi("The surface for batching is not stable yet and may change in the future.")
public class FlowController {

We are not using the CLASS named FlowController. and it is not a good idea to use a random class's setting object, because it can change anytime.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: Well, we are using BetaApi all over the place, and PubSub v1 is using this as well:
https://screenshot.googleplex.com/8KD2UYz5LETJkpi

Anyways, I am fine not to block on this. Maybe stephaniewang@ have a better insight about this.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Yeah, since the setting is simple enough. It shouldn't be a big issue either way. I just prefer to use our own definition to avoid unexpected changes in future.

They are likely to add new fields, which we don't support. It would be confusing for our users.

bigang-li (approved):

stephaniewang526 (approved):

PR #877, opened 2021-02-24
feat: StreamWriterV2 will handle schema/streamName attachment

yayi-google:
…the connection and remove stream name and schema in the following ones.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -17,6 +17,8 @@ package com.google.cloud.bigquery.storage.v1beta2;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.SettableApiFuture;
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.bigquery.storage.v1beta2.StreamConnection.DoneCallback;
 import com.google.cloud.bigquery.storage.v1beta2.StreamConnection.RequestCallback;
 import com.google.common.base.Preconditions;
@@ -24,6 +26,7 @@ import com.google.common.util.concurrent.Uninterruptibles;
 import io.grpc.Status;
 import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
+import java.io.IOException;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -322,7 +323,11 @@ private void appendLoop() {
 
       // TODO: Add reconnection here.
       while (!localQueue.isEmpty()) {
-        this.streamConnection.send(localQueue.pollFirst().message);
+        AppendRowsRequest preparedRequest =
+            prepareRequestBasedOnPosition(
+                localQueue.pollFirst().message, isFirstRequestInConnection);
+        this.streamConnection.send(preparedRequest);
+        isFirstRequestInConnection = false;
```
yirutang, line 330: Could there be a race on the isFirstRequestInConnection?
Also should the fields be cleared when after the first success response?

yayi-google: There can NOT be a race between it is a local variable accessed in a single thread. It is not shared at all.

I thought about this. If the first request failed, the entire connection will be aborted, so we don't need to wait for the first success response. In this way, we can keep it as a local variable to avoid any concurrency issue.

yirutang: I am fine with the change now, but the first failure could be a in-stream failure (where the stream is not broken, for example https://screenshot.googleplex.com/5b4zkgK89qudYDk), maybe add a comment about the potential issue here, or open a bug for me to change every failure in the first request to be a connection failure.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Discussed offline. This change may introduce some risk. So I'll spent a bit more time on this.

yirutang: Discussed offline, I think if we have client side schema compatibility check built in, the risk would be minimum (or be able to detect while client is writing their code, instead of runtime).

yayi-google: There are a lot of concerns from different people on how the schema should be passed. So I'll park this thread for now.

yayi-google: Discussed offline, we will proceed with this as step 1.

In step 2, we will accept the schema when creating StreamWriterV2 and provide a new append method.

stephaniewang526 (approved):

PR #894, opened 2021-02-26
feat: in StreamWriterV2, supports new append, which takes rows and offset

yayi-google:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -19,10 +19,12 @@ import com.google.api.core.ApiFuture;
 import com.google.api.core.SettableApiFuture;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
+import com.google.cloud.bigquery.storage.v1beta2.AppendRowsRequest.ProtoData;
 import com.google.cloud.bigquery.storage.v1beta2.StreamConnection.DoneCallback;
 import com.google.cloud.bigquery.storage.v1beta2.StreamConnection.RequestCallback;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Uninterruptibles;
+import com.google.protobuf.Int64Value;
 import io.grpc.Status;
 import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
@@ -38,8 +40,6 @@ import javax.annotation.concurrent.GuardedBy;
 
 /**
  * A BigQuery Stream Writer that can be used to write data into BigQuery Table.
- *
- * <p>TODO: Attach schema.
  *
  * <p>TODO: Attach traceId.
  *
@@ -58,6 +58,11 @@ public class StreamWriterV2 implements AutoCloseable {
    * The identifier of stream to write to.
    */
   private final String streamName;
+
+  /*
+   * The proto schema of rows to write.
+   */
+  private final ProtoSchema writerSchema;
 
   /*
    * Max allowed inflight requests in the stream. Method append is blocked at this.
@@ -135,6 +140,7 @@ public class StreamWriterV2 implements AutoCloseable {
     this.hasMessageInWaitingQueue = lock.newCondition();
     this.inflightReduced = lock.newCondition();
     this.streamName = builder.streamName;
+    this.writerSchema = builder.writerSchema;
     this.maxInflightRequests = builder.maxInflightRequest;
     this.maxInflightBytes = builder.maxInflightBytes;
     this.waitingRequestQueue = new LinkedList<AppendRequestAndResponse>();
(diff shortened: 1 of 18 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -188,10 +194,52 @@ public void run() {
...
+   * }</pre>
+   *
+   * @param rows the rows in serialized format to write to BigQuery.
+   * @param offset the offset of the first row.
+   * @return the append response wrapped in a future.
+   */
+  public ApiFuture<AppendRowsResponse> append(ProtoRows rows, long offset) {
+    // TODO: Move this check to builder after the other append is removed.
+    if (this.writerSchema == null) {
+      throw new StatusRuntimeException(
```
bigang-li, line 217: this needs to be in at the writer builder for sure, here is not a proper place to validate this.

bigang-li (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -188,10 +194,52 @@ public void run() {
...
+          Status.fromCode(Code.INVALID_ARGUMENT)
+              .withDescription("Writer schema must be provided when building this writer."));
```
stephaniewang526, line 219: If this needs to be moved then no worries but otherwise, maybe we need to add a test for this.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -59,6 +59,11 @@
    */
   private final String streamName;
 
+  /*
+   * The proto schema of rows to write.
+   */
+  private final ProtoSchema writerSchema;
```
yirutang, line 65: Could this be Descriptor? The user will need to call the conversion method (which is one of our library) in order to get to here.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -188,10 +194,52 @@ public void run() {
...
+   *     System.out.println("failed to write: " + t);
+   *   }
+   * }, MoreExecutors.directExecutor());
+   * }</pre>
+   *
+   * @param rows the rows in serialized format to write to BigQuery.
+   * @param offset the offset of the first row.
+   * @return the append response wrapped in a future.
+   */
+  public ApiFuture<AppendRowsResponse> append(ProtoRows rows, long offset) {
```
yirutang, line 214: Should we simply accept byte strings?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: +1

yayi-google: Sure. That's why I have a TODO here.

I need to put it here instead for backward compatibility reason. Otherwise, it would break existing customers. For example, the sample will stop working because the version is bumped up automatically.

I just need to do it in two phases.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: I think it is similar either way. And we already agree on this in the last discussion. I would like to avoid go back and get everybody agree on a new thing given it is similar either way.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Same as above.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: This is not schema compact check. I just want to bring it up. I am fine keeping this interface more open and leave options to users.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: I think byte string is cleaner in this case. But since this builds with API anyway, I am fine with ProtoRows.

yirutang (approved):

stephaniewang526 (approved):

PR #895, opened 2021-02-26
feat: update StreamWriterV2 to support trace id

yayi-google:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -137,6 +140,7 @@ private StreamWriterV2(Builder builder) throws IOException {
     this.streamName = builder.streamName;
     this.maxInflightRequests = builder.maxInflightRequest;
     this.maxInflightBytes = builder.maxInflightBytes;
+    this.traceId = builder.traceId;
```
bigang-li, line 143: let's start with the user trace id/job id here?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
```diff
@@ -531,6 +540,12 @@ public Builder setCredentialsProvider(CredentialsProvider credentialsProvider) {
       return this;
     }
 
+    /** TraceId for debuging purpose. */
+    public Builder setTraceId(String traceId) {
+      this.traceId = traceId;
+      return this;
```
bigang-li, line 546: Let's perform some simple sanity check here.

On the server side, this needs to be something like key value pair like http header or A:B; is okay as well (we need to document it).

A simple processing at the client side will be splitting the trace id per the standard above if we can;
other wise convert it to UserTrace:xxx;

On top of it, adding the StreamWrite:Version?

stephaniewang526:
Can you take a look at this?
```
StreamWriterTest.testFlushAllFailed:893 Unexpected exception:java.lang.InterruptedException: Wait timeout
```
I have seen this happen a few times now. I worry it's a flaky test.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: It is supposed to be general, so I don't want to limit it to job id. I feel like trace id is a more general term here.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: All these requires backend support, which is not there yet.

I would also prefer to put all the logic you mentioned into the server side, because we have strong control of the codes there.

On the other hand, it is very hard to change the client library. So a free string seems most flexible.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: This will be used for all use cases, not just dataflow. So I would prefer to more general term like trace id here.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yirutang: Making dataflow do the right thing is also fine to me. Just that if they mess up, then we will have issues in debugging their traffic. The backend already parses these trace ids.

yirutang (approved):

yayi-google:
> Can you take a look at this?
> 
> ```
> StreamWriterTest.testFlushAllFailed:893 Unexpected exception:java.lang.InterruptedException: Wait timeout
> ```
> 
> I have seen this happen a few times now. I worry it's a flaky test.
> [Full trace](https://github.com/googleapis/java-bigquerystorage/pull/895/checks?check_run_id=1983959190)

Taken care in a separate PR.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
bigang-li: Yes, we need to perform such kind of logic in the server side, but here, let's perform some sanity check and makes sure that final trace id is in a good shape as well?

stephaniewang526:
Looks like there are merge conflicts -- PTAL.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
yayi-google: Done. Added check to make sure traceId follow the format of A:B.

yayi-google:
> Looks like there are merge conflicts -- PTAL.

Done. Fixed the conflicts.

stephaniewang526 (approved):

stephaniewang526 (approved):

PR #940, opened 2021-03-15
docs(samples): Check for error from BatchCommitWriteStreams

MikeWasson:
If the returned BatchCommitWriteStreamsResponse does not have a commit time, it means an error occurred.

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
+++ b/samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
@@ -90,6 +90,10 @@ public class WritePendingStream {
               .build();
       BatchCommitWriteStreamsResponse commitResponse =
           client.batchCommitWriteStreams(commitRequest);
+      // If the response does not have a commit time, it means the commit operation failed.
+      if (commitResponse.hasCommitTime() == false) {
+        throw new RuntimeException("Error committing the streams");
+      }
       System.out.println("Appended and committed records successfully.");
     } catch (ExecutionException e) {
       // If the wrapped exception is a StatusRuntimeException, check the state of the operation.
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
```diff
@@ -90,6 +90,10 @@ public static void writePendingStream(String projectId, String datasetName, Stri
               .build();
       BatchCommitWriteStreamsResponse commitResponse =
           client.batchCommitWriteStreams(commitRequest);
+      // If the response does not have a commit time, it means the commit operation failed.
+      if (commitResponse.hasCommitTime() == false) {
```
stephaniewang526, line 94: Does it not return an error object in this case? User is expected to guess that something went wrong?

MikeWasson: The user can call getStreamErrorsList to get a list of StorageError's. You could use this to look at the error code per stream. Does it make sense to show this?

stephaniewang526: I think it would make more sense to just show the error objects rather than show hasCommitTime indirectly.

MikeWasson: Per @yirutang checking for hasCommitTime is more semantically meaningful, because that's the error condition. I added code to display the list of error messages.

stephaniewang526: Got it. SGTM.

stephaniewang526 (approved):

PR #971, opened 2021-03-29
feat: BigDecimal and ByteString encoding

JacobStocklass:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This code was ported from ZetaSQL and can be found here:
+ * https://github.com/google/zetasql/blob/c55f967a5ae35b476437210c529691d8a73f5507/java/com/google/zetasql/Value.java
+ */
+
+package com.google.cloud.bigquery.storage.v1beta2;
+
+import com.google.common.primitives.Bytes;
+import com.google.protobuf.ByteString;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+public class BigDecimalByteStringEncoder {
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+          "-578960446186580977117854925043439539266.34992332820282019728792003956564819968");
+
+  public static BigDecimal encodeToBigDecimal(ByteString byteString) {
+    BigDecimal bigDecimal =
+        deserializeBigDecimal(
+            byteString, scale, MAX_BIGNUMERIC_VALUE, MIN_BIGNUMERIC_VALUE, "BigDecimal");
+    return bigDecimal;
+  }
+
+  public static ByteString encodeToByteString(BigDecimal bigDecimal) {
```
yirutang, line 45: The name should mention BigQuery NUMERIC type, we have another type called BIGNUMERIC which needs to be supported separately.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+public class BigDecimalByteStringEncoder {
+  private static int scale = 38;
+  private static final BigDecimal MAX_BIGNUMERIC_VALUE =
+      new BigDecimal(
+          "578960446186580977117854925043439539266.34992332820282019728792003956564819967");
+  private static final BigDecimal MIN_BIGNUMERIC_VALUE =
+      new BigDecimal(
+          "-578960446186580977117854925043439539266.34992332820282019728792003956564819968");
+
+  public static BigDecimal encodeToBigDecimal(ByteString byteString) {
```
yirutang, line 38: decode?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
JacobStocklass: Okay that should be easy. Should be the same but with a different scale, max and min

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
JacobStocklass: I can rename to decode so that they line up better

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/BigDecimalByteStringEncoder.java
JacobStocklass: Renamed method

JacobStocklass:
> NumericScale

My understanding is that the BigNumeric scale is 38 and the NumericScale is 9 with different Max and Min

yirutang (approved):

stephaniewang526 (approved):

PR #979, opened 2021-03-31
test: BigDecmialByteStringConverter E2E Test

JacobStocklass:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -238,12 +238,15 @@ public class JsonToProtoMessage {
           break;
         case BYTES:
           if (val instanceof String) {
+            // TODO(jstocklass): If string, decode it and pass in the byte array. Will need to update
+            // tests to ensure that strings passed in are properly encoded as well.
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
           } else if (val instanceof JSONArray) {
             try {
               byte[] bytes = new byte[((JSONArray) val).length()];
-              for (int j = 0; j < ((JSONArray) val).length(); j++) {
-                bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
+                for (int j = 0; j < ((JSONArray) val).length(); j++) {
+                  bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
+                // TODO(jstocklass): Add a unit test for this
               }
               protoMsg.addRepeatedField(fieldDescriptor, bytes);
             } catch (ClassCastException e) {
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -252,7 +252,11 @@ public class JsonToProtoMessage {
             } catch (ClassCastException e) {
               throw new IllegalArgumentException(
                   String.format(
-                      "Error: " + currentScope + "["+ index + "] could not be converted to byte[]."));
+                      "Error: "
+                          + currentScope
+                          + "["
+                          + index
+                          + "] could not be converted to byte[]."));
             }
           } else {
             fail = true;
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -239,8 +239,7 @@ public class JsonToProtoMessage {
         case BYTES:
           if (val instanceof String) {
             // TODO(jstocklass): If string, decode it and pass in the byte array. Will need to
-            // update
-            // tests to ensure that strings passed in are properly encoded as well.
+            // update tests to ensure that strings passed in are properly encoded as well.
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
           } else if (val instanceof JSONArray) {
             try {
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -247,11 +247,12 @@ public class JsonToProtoMessage {
               byte[] bytes = new byte[((JSONArray) val).length()];
               for (int j = 0; j < ((JSONArray) val).length(); j++) {
                 bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
-                // TODO(jstocklass): Add a unit test for this
               }
               protoMsg.addRepeatedField(fieldDescriptor, bytes);
             } catch (ClassCastException e) {
-              fail = true;
+              throw new IllegalArgumentException(
+                  String.format(
+                      "Error: " + currentScope + "["+ index + "] could not be converted to byte[]."));
             }
           } else {
             fail = true;
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -240,8 +240,6 @@ public class JsonToProtoMessage {
           if (val instanceof String) {
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
           } else if (val instanceof JSONArray) {
-            // TODO: what if this JSONArray is full of some other weird stuff...
-            // Like a jsonarray of strings or something this might get wrecked
             try {
               byte[] bytes = new byte[((JSONArray) val).length()];
               for (int j = 0; j < ((JSONArray) val).length(); j++) {
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -238,14 +238,15 @@ public class JsonToProtoMessage {
           break;
         case BYTES:
           if (val instanceof String) {
-            // TODO(jstocklass): If string, decode it and pass in the byte array. Will need to update
+            // TODO(jstocklass): If string, decode it and pass in the byte array. Will need to
+            // update
             // tests to ensure that strings passed in are properly encoded as well.
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
           } else if (val instanceof JSONArray) {
             try {
               byte[] bytes = new byte[((JSONArray) val).length()];
-                for (int j = 0; j < ((JSONArray) val).length(); j++) {
-                  bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
+              for (int j = 0; j < ((JSONArray) val).length(); j++) {
+                bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
                 // TODO(jstocklass): Add a unit test for this
               }
               protoMsg.addRepeatedField(fieldDescriptor, bytes);
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigquery.storage.v1beta2;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.ByteString;
 import com.google.protobuf.Descriptors.Descriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.DynamicMessage;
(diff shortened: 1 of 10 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -138,7 +139,10 @@ private static void fillField(
         }
         break;
       case BYTES:
-        if (val instanceof String) {
+        if (val instanceof ByteString) {
+          protoMsg.setField(fieldDescriptor, ((ByteString) val).toByteArray());
+          return;
+        } else if (val instanceof String) {
           protoMsg.setField(fieldDescriptor, ((String) val).getBytes());
```
yirutang, line 146: If it is a string, we want to decoded it and pass in the byte array. It is well known that bytes needs to be encoded in order to be put into string. I don't think ((String) val).getBytes()) will work for bytes. It needs to be fixed. You can fix it in a separate PR.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -235,7 +239,14 @@ private static void fillRepeatedField(
         case BYTES:
           if (val instanceof String) {
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
```
yirutang, line 241: Does this path work?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/it/ITBigQueryBigDecimalByteStringEncoderTest.java
```diff
@@ -0,0 +1,162 @@
...
+    DatasetInfo datasetInfo =
+        DatasetInfo.newBuilder(/* datasetId = */ DATASET).setDescription(DESCRIPTION).build();
+    bigquery.create(datasetInfo);
+    tableInfo =
+        TableInfo.newBuilder(
+                TableId.of(DATASET, TABLE),
+                StandardTableDefinition.of(
+                    Schema.of(
+                        com.google.cloud.bigquery.Field.newBuilder(
+                                "test_str", StandardSQLTypeName.STRING)
```
yirutang, line 78: We don't need this test.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -235,7 +239,14 @@ private static void fillRepeatedField(
...
+          } else if (val instanceof JSONArray) {
```
yirutang, line 242: A byte array becomes a type of JSON Array is a bit strange. In your case, you happen to have a byte array in the JSON array, but what if you don't have byte array in the json array?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
JacobStocklass: I will work on that in another PR, I think that this is how it was originally set up and I only added the ByteString version. I think that you're right and it should be changed though. This will require some other changes to make sure that our tests don't break.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/it/ITBigQueryBigDecimalByteStringEncoderTest.java
JacobStocklass: Deleted

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
JacobStocklass: That's a good point, I will add a check for the array being full of bytes and if it isn't maybe we should have a unique error message thrown. I'll also add a test that expects this error in the JsonToProtoMessageTest file.

JacobStocklass: I made the check so that if anything in the array doesn't behave nicely, or convert to a byte without giving an error, the default error message is thrown stating that the instance does not have the correct type. This is now covered in the JsonToProtoMessageTest as well.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
yirutang: Just make sure it is addressed in the other PR.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
yirutang: Got it. SGTM.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -235,6 +239,16 @@ private static void fillRepeatedField(
...
+            try {
+              byte[] bytes = new byte[((JSONArray) val).length()];
+              for (int j = 0; j < ((JSONArray) val).length(); j++) {
+                bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
```
yirutang, line 246: We should add a unit test for this.

JacobStocklass: I've added a test, adding it in to the existing test proved very complicated since it required a repeated set of BYTE insertions. I just made a similar test that stands alone to cover these lines.

JacobStocklass:
@yirutang PTAL

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/BQTableSchemaToProtoDescriptorTest.java
```diff
@@ -183,7 +184,7 @@ public void testStructComplex() throws Exception {
             .build();
     final Table.TableFieldSchema TEST_NUMERIC =
         Table.TableFieldSchema.newBuilder()
-            .setType(Table.TableFieldSchema.Type.NUMERIC)
```
yirutang, line 186: Why change this to BYTES? It is called TEST_NUMERIC.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -234,7 +238,21 @@ private static void fillRepeatedField(
...
+            // TODO(jstocklass): If string, decode it and pass in the byte array. Will need to
+            // update
+            // tests to ensure that strings passed in are properly encoded as well.
             protoMsg.addRepeatedField(fieldDescriptor, ((String) val).getBytes());
+          } else if (val instanceof JSONArray) {
+            try {
+              byte[] bytes = new byte[((JSONArray) val).length()];
+              for (int j = 0; j < ((JSONArray) val).length(); j++) {
+                bytes[j] = (byte) ((byte) (((JSONArray) val).get(j)) & 0xFF);
+                // TODO(jstocklass): Add a unit test for this
```
yirutang, line 250: Add negative tests.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
```diff
@@ -234,7 +238,21 @@ private static void fillRepeatedField(
...
+              }
+              protoMsg.addRepeatedField(fieldDescriptor, bytes);
+            } catch (ClassCastException e) {
+              fail = true;
```
yirutang, line 254: Would be nice to have a detailed error message. What does the message currently look like?

yirutang (commented):
Looks good, some comments here.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/BQTableSchemaToProtoDescriptorTest.java
JacobStocklass: I changed it here just to avoid having to also change the v1alpha2 JstonToProtoMessage and other supporting changes. This test references the proto/jsonTest.proto file that I changed in order to test the changes made in v1beta2. This just sets the type here to byte since we changed the numeric to be also be bytes. This just makes the test pass so that maven doesn't get angry

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
JacobStocklass: I'll change the error message. How does something like "Error: root.test_repeated[0] could not be converted to byte[]." sound?

yirutang: SG.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1alpha2/BQTableSchemaToProtoDescriptorTest.java
yirutang: nvm

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessage.java
JacobStocklass: I've added a negative test as well now, it covers the failure if the json cannot be converted to bytes and we get the new error message.

JacobStocklass:
@yirutang PTAL

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
```diff
@@ -108,6 +117,21 @@
           .put(RepeatedObject.getDescriptor(), "object")
           .build();
 
+  private static ImmutableMap<Descriptor, Message[]> BytesRepeatedCorrectProto =
```
yirutang, line 120: Could you add this case directly to AllTypesToDebugMessageTest?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
```diff
@@ -345,8 +391,32 @@ public void testAllTypes() throws Exception {
...
+        assertEquals(1, success);
+      } else {
+        assertEquals(0, success);
+      }
+    }
+  }
+
   @Test
   public void testAllRepeatedTypesWithLimits() throws Exception {
+    // Mark1
```
yirutang, line 419: ?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
```diff
@@ -345,8 +391,32 @@ public void testAllTypes() throws Exception {
...
+          DynamicMessage protoMsg =
+              JsonToProtoMessage.convertJsonToProtoMessage(entry.getKey(), json);
+          assertEquals(protoMsg, BytesRepeatedCorrectProto.get(entry.getKey())[success]);
+          success += 1;
+        } catch (IllegalArgumentException e) {
+          assertEquals(
+              "Error: root.test_repeated[0] could not be converted to byte[].", e.getMessage());
+        }
+      }
+      if (entry.getKey() == RepeatedBytes.getDescriptor()) {
```
yirutang, line 409: does this test have 2 cases? I only see one entry in the ByteRepeatedTypesToDebugMessageTest. It is quite hard to read. Could you just add the positive case to the AllTypesToDebugMessageTest and then focus on the negative test here?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
JacobStocklass: I've been working on adding it in like that but the issue is it runs into an error with multiple entries with the same key because of the existing BYTES related test that covers the String path. I've been messing with adding it in and it's resulting in some very messy nesting and checking against multiple possible outcomes that obscures what is being done and how it fails. I can probably make it work but I don't know if it is the best option.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
JacobStocklass: Oops, that was just so I could ctrl+f around the code easier

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
JacobStocklass: I was messing with that for a while but I think your comment below actually kind of solves the issue I was having with it. I just added in the positive test to the AllRepeatedTypesToDebugmessageTest and it works now. I will see if the negative test can also be added in now.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1beta2/JsonToProtoMessageTest.java
JacobStocklass: I realized that there was a better way to add in both of the positive and negative tests to the existing test. It should be covered now and much easier to understand what changes I made

yirutang (approved):

stephaniewang526 (approved):

PR #984, opened 2021-04-07
fix: Cleanup JsonWriter bytes conversion code and add some test coverage

yirutang:


stephaniewang526 (approved):

stephaniewang526:
Some tests after failing. PTAL @yirutang

stephaniewang526 (changes requested):
Some tests after failing. PTAL @yirutang

JacobStocklass (approved):
This seems a lot better with Bytes being handled as just JSONArray instead of JSONArray or String like I had it earlier. Makes the testing a lot less confusing.

stephaniewang526 (approved):

PR #1028, opened 2021-04-22
fix: Add OnSchemaUpdateRunnable to StreamWriterV2, so that JsonWriter…

yirutang:
… can switch over to use V2.

This is needed in order to switch JSON writer to use StreamWriterV2.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/StreamWriterV2.java
@@ -210,9 +210,11 @@ public class StreamWriterV2 implements AutoCloseable {
               }
             });
     this.appendThread.start();
-    executor = InstantiatingExecutorProvider.newBuilder()
-                   .setExecutorThreadCount(Runtime.getRuntime().availableProcessors())
-                   .build().getExecutor();
+    executor =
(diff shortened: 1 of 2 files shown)


yayi-google (commented):
I don't think this aligns with what we agreed on schema change.

I would rather not have such function than introducing hard to handle race conditions.

I think we should just fail the schema if schema changes for now.

yirutang:
> I don't think this aligns with what we agreed on schema change.
> 
> I would rather not have such function than introducing hard to handle race conditions.
> 
> I think we should just fail the schema if schema changes for now.

The reason for the change is to migrate JsonWriter over to StreamWriterV2, JsonWriter supports schema update.

In is runnable, it will call stream reconnect. I think our stream reconnect is thread safe, and runnable is scheduled to run on some other thread, so we don't need to worry about race here?

yayi-google:
> > I don't think this aligns with what we agreed on schema change.
> > I would rather not have such function than introducing hard to handle race conditions.
> > I think we should just fail the schema if schema changes for now.
> 
> The reason for the change is to migrate JsonWriter over to StreamWriterV2, JsonWriter supports schema update.
> 
> In is runnable, it will call stream reconnect. I think our stream reconnect/recreate is thread safe, and runnable is scheduled to run on some other thread, so we don't need to worry about race here?

This is the wrong reason to add such unsafe feature. You should remove schema update support in JsonWriter if you want to migrate.

PR #1058, opened 2021-05-07
fix: migrate json writer to use StreamWriterV2

yirutang:


google-cloud-bigquerystorage/clirr-ignored-differences.xml
```diff
@@ -3,27 +3,8 @@
 <differences>
     <!-- Allow below protobuf changes as non-breaking-->
     <difference>
-        <differenceType>8001</differenceType>
```
stephaniewang526, line 6: please don't remove any of these existing `<difference>`s since they are needed to allow protobuf changes to be non-breaking.

stephaniewang526 (changes requested):

stephaniewang526 (approved):

PR #1155, opened 2021-06-29
docs: user-provided streamOrTableName for default write streams

anguillanneuf:
Fixes a broken URL in a sample comment.

Also cleans up the docstrings for `streamOrTableName` in `JsonStreamWriter.Builder`.

IIUC, `createDefaultStream` can only be set to True on `StreamWriter` not `JsonStreamWriter`, the language below could mislead devs into thinking otherwise because it's not clear which `builder` it refers to.  
https://github.com/googleapis/java-bigquerystorage/blob/d6c3146d21282d1e0724583f68a31184d2d7167b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonStreamWriter.java#L185-L188

Also it looks like `_default` is not added by client code but by the library.
https://github.com/googleapis/java-bigquerystorage/blob/d6c3146d21282d1e0724583f68a31184d2d7167b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonStreamWriter.java#L267-L274

Not sure if I'm making all the right calls here. Thanks for reviewing!

---
Also see internal cl/381977832.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1beta2/JsonStreamWriter.java
@@ -183,8 +183,8 @@ public class JsonStreamWriter implements AutoCloseable {
    * StreamWriter by default.
    *
    * @param streamOrTableName name of the stream that must follow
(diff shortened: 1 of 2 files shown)


stephaniewang526 (approved):

stephaniewang526:
Thank you!

PR #1288, opened 2021-09-03
fix: Accept null json values in JsonToProtoMessage converter

### LINKED ISSUE #1258: JsonToProtoMessage.fillField can not handle null values
JsonStreamWriter.append() fails when adding a JSONArray in which some fields are null because JsonToProtoMessage.fillField() does not handle null values returned by java.lang.Object val = json.get(exactJsonKeyName); (line 135). 

null values should be supported because the Schema used to create the JsonStreamWriter could contain fields with a  NULLABLE mode.

yirutang:
Fixes #1258 ☕️


stephaniewang526 (approved):

stephaniewang526:
could you see if the test case needs to be updated?
```
Error:  Errors: 
Error:    StreamWriterTest.testFlowControlBehaviorBlockAbortOnShutdown:984 » NullPointer
```
[Full trace](https://github.com/googleapis/java-bigquerystorage/pull/1288/checks?check_run_id=3501768492#step:6:5659)

stephaniewang526 (approved):

PR #1338, opened 2021-09-30
docs(samples): Add WriteAPI BUFFERED mode sample

VeronicaWasson:
Adds a sample for using the JsonStreamWriter in buffered mode.


samples/snippets/src/test/java/com/example/bigquerystorage/WriteBufferedStreamIT.java
```diff
@@ -0,0 +1,96 @@
...
+  @After
+  public void tearDown() {
+    bigquery.delete(
+        DatasetId.of(GOOGLE_CLOUD_PROJECT, datasetName), DatasetDeleteOption.deleteContents());
+    System.setOut(null);
+  }
+
+  @Test
+  public void testWriteBufferedStream() throws Exception {
+    WriteBufferedStream.writeBufferedStream(GOOGLE_CLOUD_PROJECT, datasetName, tableName);
```
yirutang, line 93: Could you add a query to verify rows are there?

samples/snippets/src/main/java/com/example/bigquerystorage/WriteBufferedStream.java
```diff
@@ -0,0 +1,100 @@
...
+          AppendRowsResponse response = future.get();
+        }
+
+        // Flush the buffer.
+        FlushRowsRequest flushRowsRequest =
+            FlushRowsRequest.newBuilder()
+                .setWriteStream(writeStream.getName())
+                .setOffset(Int64Value.of(10 * 2 - 1)) // Advance the cursor to the latest record.
+                .build();
+        FlushRowsResponse flushRowsResponse = client.flushRows(flushRowsRequest);
```
yirutang, line 88: Maybe add another write without flush and comment that those rows will not appear in the table.

samples/snippets/src/test/java/com/example/bigquerystorage/WriteBufferedStreamIT.java
VeronicaWasson: Added

samples/snippets/src/main/java/com/example/bigquerystorage/WriteBufferedStream.java
VeronicaWasson: IMO this would some unneeded complexity to the sample, and also make it harder to compare the differences with pending and committed modes. Also it would make the code not strictly correct, if a user didn't read the code comments carefully.

yirutang (approved):

stephaniewang526 (approved):

PR #1357, opened 2021-10-13
Fix decimal mapping when value is array

### LINKED ISSUE #1356: BigQuery storage: errors while trying to insert bytes array
I've encountered this, while trying to insert BigDecimal value into BQ (NUMERIC column). I converted my decimal value to byte array and tried to insert it, but was greeted with exception:
```
IllegalArgumentException: JSONObject does not have a bytes field at root.col_numeric.
```

I've dug into the source and discovered that `JsonToProtoMessage` is missing return statement in cases when `fieldDescriptor.getType` has BYTES and value comes as instance of `JSONArray`.

Note: I could not use string representation of BigDecimal and let `BigDecimalByteStringEncoder` do it's work, cause I am forced to use lower version of guava. The class mentioned uses a `Bytes.reverse` method, which is not present in guava 20.0.

#### Environment details

1. BigQuery storage library
2. OS type and version: macOs Big Sur Version 11.4
3. Java version: 1.8.0_232
4. bigquerystorage version(s): 2.3.3

#### Steps to reproduce

  1. Have a table in BQ with column type NUMERIC
  2. Run adapted example code (i'm sorry, it's in Scala)

#### Code example

```scala
val obj = new JSONObject()
val bytes = BigDecimal(1.9).bigDecimal.setScale(9).unscaledValue().toByteArray.reverse
// or just: val bytes = Seq(0, -77, 63, 113).toArray

obj.put("col_numeric", new JSONArray(bytes))
val objects = Seq(obj).asJava.toArray

val array: JSONArray = new JSONArray(objects)
val future = writer.append(array)
future.get()
```

#### Stack trace
```
Exception in thread "main" java.lang.IllegalArgumentException: JSONObject does not have a bytes field at root.col_numeric.
	at com.google.cloud.bigquery.storage.v1beta2.JsonToProtoMessage.fillField(JsonToProtoMessage.java:307)
	at com.google.cloud.bigquery.storage.v1beta2.JsonToProtoMessage.convertJsonToProtoMessageImpl(JsonToProtoMessage.java:139)
	at com.google.cloud.bigquery.storage.v1beta2.JsonToProtoMessage.convertJsonToProtoMessage(JsonToProtoMessage.java:87)
	at com.google.cloud.bigquery.storage.v1beta2.JsonStreamWriter.append(JsonStreamWriter.java:110)
	at com.google.cloud.bigquery.storage.v1beta2.JsonStreamWriter.append(JsonStreamWriter.java:90)
```

ramasLTU:
Added missing return, when processing big decimal field which has its value in byte array format.
Previously algorithm would do the mapping, set the values, but would not exit with return and fallback to IllegalArgumentException at the end of the method.

Fixes #1356☕️

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,13 @@ implementation 'com.google.cloud:google-cloud-bigquerystorage'
 If you are using Gradle without BOM, add this to your dependencies
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-bigquerystorage:2.3.3'
+implementation 'com.google.cloud:google-cloud-bigquerystorage:2.4.0'
 ```
 
 If you are using SBT, add this to your dependencies
 
 ```Scala
-libraryDependencies += "com.google.cloud" % "google-cloud-bigquerystorage" % "2.3.3"
+libraryDependencies += "com.google.cloud" % "google-cloud-bigquerystorage" % "2.4.0"
 ```
 
 ## Authentication


ramasLTU:
@googlebot I signed it!

yirutang (approved):

stephaniewang526 (approved):

PR #1358, opened 2021-10-13
Make client robust to Rst from OkHTTP

emkornfield:
The
[OKHTTP](https://github.com/grpc/grpc-java/blob/d4fa0ecc07495097453b0a2848765f076b9e714c/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java#L1241)
transport uses slightly different error messages for RST Stream that
aren't caught by current screen scraping.  This adds a check for the
strict explicitly.

bug: Retry RST Stream on OKHTTP.


emkornfield:
CC @stephaniewang526

PR #1369, opened 2021-10-19
fix: BQ/proto schema names should be compared lowercase

### LINKED ISSUE #1366: BigQuery: JSON writer does not function if table schema has fields with mixed-case names
Thanks for stopping by to let us know something could be better!

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ..."). 
   General, Core, and Other are also allowed as types
2. OS type and version: N/A
3. Java version: N/A
4. bigquerystorage version(s): `2.4.0` (using both `v1` and `v1beta2` API versions)

#### Steps to reproduce

  1. Create a bigquery table with a field with mixed case, such as `fooBar`
  2. Attempt to use `JsonStreamWriter` to save a record to this table

#### Stack trace
```
com.google.api.pathtemplate.ValidationException: Field at index 3 has mismatch names (publishedAt) (publishedat)
    at com.google.cloud.bigquery.storage.v1beta2.JsonToProtoMessage.convertJsonToProtoMessageImpl(JsonToProtoMessage.java:133)
    at com.google.cloud.bigquery.storage.v1beta2.JsonToProtoMessage.convertJsonToProtoMessage(JsonToProtoMessage.java:86)
    at com.google.cloud.bigquery.storage.v1beta2.JsonStreamWriter.append(JsonStreamWriter.java:110)
    at com.google.cloud.bigquery.storage.v1beta2.JsonStreamWriter.append(JsonStreamWriter.java:90)
```

This is because the comparison here does not take into account the fact that the protobuf schema always has lower case field names, while the the BQ table schema field may have mixed-cased field names:
https://github.com/googleapis/java-bigquerystorage/blob/9d272dd711dfe8909a7b421dabdcdef30f1dc689/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java#L127

stephenjudkins:

Fixes #1366☕️

### DIFF
--- a/.kokoro/dependencies.sh
+++ b/.kokoro/dependencies.sh
@@ -42,7 +42,6 @@ function determineMavenOpts() {
     then
       # MaxPermSize is no longer supported as of jdk 17
       echo -n "-Xmx1024m"
-      ;;
     else
       echo -n "-Xmx1024m -XX:MaxPermSize=128m"
     fi


yirutang (commented):
Thanks for fixing this, looks good. Could you add a test case to capture this change?

yirutang (approved):

stephaniewang526 (changes requested):
Please fix lint failure by running `mvn com.coveo:fmt-maven-plugin:format` and pushing the change

PR #1391, opened 2021-10-29
feat: add support for StorageError

stephaniewang526:
Fixes b/203003176


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -482,6 +489,10 @@ private void doneCallback(Throwable finalStatus) {
     } finally {
       this.lock.unlock();
     }
+    Exceptions.StorageException storageException = Exceptions.toStorageException(finalStatus);
+    if (storageException != null) {
+      throw new StorageException(finalStatus);
```
yirutang, line 494: Just set the connectionFinalStatus to be the upgraded stroageException like before.

stephaniewang526: Done

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/StreamWriterTest.java
```diff
@@ -304,6 +306,38 @@ public void testAppendSuccessAndInStreamError() throws Exception {
     writer.close();
   }
 
+  @Test
```
yirutang, line 309: Add another test case on the new error handling you added on onDone.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+    return null;
+  }
+
+  /* Converts a c.g.rpc.Status into a StorageException, if possible.
+   * Examines the embedded StorageError, and potentially returns a StreamFinalizedException or
+   * SchemaMismatchedException (both derive from StorageException).
+   * If there is no StorageError, or the StorageError is a different error it will return NULL.
+   */
```
BenWhitehead, line 104: Javadoc comments start with `/**`
```suggestion
  /**
   * Converts a c.g.rpc.Status into a StorageException, if possible. Examines the embedded
   * StorageError, and potentially returns a {@link StreamFinalizedException} or {@link
   * SchemaMismatchedException} (both derive from StorageException). If there is no StorageError, or
   * the StorageError is a different error it will return NULL.
   */
```

I've also added javadoc links to the classes that can be returned for convenience.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      default:
+        return null;
+    }
+  }
+
+  /* Converts a Throwable into a StorageException, if possible.
+   * Examines the embedded error message, and potentially returns a StreamFinalizedException or
+   * SchemaMismatchedException (both derive from StorageException).
+   * If there is no StorageError, or the StorageError is a different error it will return NULL.
+   */
```
BenWhitehead, line 128: Ditto

```suggestion
  /**
   * Converts a Throwable into a StorageException, if possible. Examines the embedded error message,
   * and potentially returns a {@link StreamFinalizedException} or {@link SchemaMismatchedException}
   * (both derive from StorageException). If there is no StorageError, or the StorageError is a
   * different error it will return NULL.
   */
```

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      super(message, cause);
+      setStreamName(name);
+    }
+  }
+
+  /**
+   * There was a schema mismatch due to bigquery table with fewer fields than the input message.
+   * This can be resolved by updating the table's schema with the message schema.
+   */
+  public static class SchemaMismatchedException extends StorageException {
```
BenWhitehead, line 78: This class should be final unless we have a specific reason for it not to be.
```suggestion
  public static final class SchemaMismatchedException extends StorageException {
```

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      return streamName;
+    }
+
+    protected void setStreamName(String value) {
+      this.streamName = value;
+    }
+  }
+
+  /** Stream has already been finalized. */
+  public static class StreamFinalizedException extends StorageException {
```
BenWhitehead, line 65: This class should be final unless we have a specific reason for it not to be.

```suggestion
  public static final class StreamFinalizedException extends StorageException {
```

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+    protected StreamFinalizedException() {}
```
BenWhitehead, line 66: I think this default constructor can be removed since it's not used by either of the factory methods defined in the wrapper class.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      setStreamName(name);
+    }
+  }
+
+  /**
+   * There was a schema mismatch due to bigquery table with fewer fields than the input message.
+   * This can be resolved by updating the table's schema with the message schema.
+   */
+  public static class SchemaMismatchedException extends StorageException {
+    protected SchemaMismatchedException() {}
```
BenWhitehead, line 79: I think this default constructor can be removed since it's not used by either of the factory methods defined in the wrapper class.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+import com.google.api.gax.grpc.GrpcStatusCode;
+import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.Any;
+import com.google.protobuf.InvalidProtocolBufferException;
+import javax.annotation.Nullable;
+
+/** Exceptions for Storage Client Libraries. */
+public final class Exceptions {
+  /** Main Storage Exception. Might contain map of streams to errors for that stream. */
+  public static class StorageException extends RuntimeException {
```
BenWhitehead, line 27: I think we want to make this class effectively final, by making all of the constructors private so that even if someone attempts to extend this class they won't be able to create a constructor thereby getting is a "pseudo-sealed" class since only classes in this file will have access to the private constructors.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,145 @@
...
+      this.errors = value;
+    }
+
+    private String streamName;
+
+    public String getStreamName() {
+      return streamName;
+    }
+
+    protected void setStreamName(String value) {
```
BenWhitehead, line 59: I'd recommend removing these setters, and instead moving them to the constructor. This allows us to make this exception immutable which makes it safer and less foot-gun prone for our customers.

From a little poking I was able to come up with this minimized immutable base class which shows what I think we can do to trim things down, and make this a sealed and immutable. (I've also also added some more annotations for `@Nullable`)

```java
  public static class StorageException extends RuntimeException {

    private final ImmutableMap<String, GrpcStatusCode> errors;

    private StorageException() {
      this(null, null, null, ImmutableMap.of());
    }

    private StorageException(
        @Nullable String message,
        @Nullable Throwable cause,
        @Nullable String streamName,
        ImmutableMap<String, GrpcStatusCode> errors) {
      super(message, cause);
      checkArgument(errors != null, "errors must be non null");
      this.streamName = streamName;
      this.errors = errors;
    }

    public ImmutableMap<String, GrpcStatusCode> getErrors() {
      return errors;
    }

    @Nullable
    public String getStreamName() {
      return streamName;
    }
  }
```

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/StreamWriterTest.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Thanks! Updated.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Thanks! Updated.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done. Changes in commit #8a2f95f

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public String getStreamName() {
+      return streamName;
+    }
+  }
+
+  /** Stream has already been finalized. */
+  public static final class StreamFinalizedException extends StorageException {
+    protected StreamFinalizedException(String name, String message, Throwable cause) {
+      super(message, cause, name, null);
```
BenWhitehead, line 61: This will fail the null check in the constructor, will need to provide an empty collection instead.
```suggestion
      super(message, cause, name, ImmutableMap.of());
```

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  /**
+   * There was a schema mismatch due to bigquery table with fewer fields than the input message.
+   * This can be resolved by updating the table's schema with the message schema.
+   */
+  public static final class SchemaMismatchedException extends StorageException {
+    protected SchemaMismatchedException(String name, String message, Throwable cause) {
+      super(message, cause, name, null);
```
BenWhitehead, line 71: This will fail the null check in the constructor, will need to provide an empty collection instead.
```suggestion
      super(message, cause, name, ImmutableMap.of());
```

BenWhitehead (changes requested):
Two bugs, and the reason your tests builds are all hung and will likely need to timeout.

An exception is being thrown because the map is null while trying to initialize the exception most likely going into a loop or something.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * Converts a Throwable into a StorageException, if possible. Examines the embedded error message,
+   * and potentially returns a {@link StreamFinalizedException} or {@link SchemaMismatchedException}
+   * (both derive from StorageException). If there is no StorageError, or the StorageError is a
+   * different error it will return NULL.
+   */
+  @Nullable
+  public static StorageException toStorageException(Throwable exception) {
```
yirutang, line 120: Is there a way to convert the throwable back to a rpcStatus, I just realized that this is a string match.

stephaniewang526: What type of Throwable do you have in mind? The problem is the Throwable only provides `getMesasge()` that contains information that specifically relates to `STREAM_FINALIZED` and `SCHEMA_MISMATCH_EXTRA_FIELDS` errors (unlike `AppendRowsResponse` which packs in error code and details). I don't think we can get away from the String match.

BenWhitehead (approved):

yirutang (approved):
Thanks!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Added a few TODO items to follow up when cl/408735437 is rolled out.

Neenu1995 (approved):

steffnay (approved):

PR #1397, opened 2021-11-03
feat: DATE and DOUBLE support more input formats

### LINKED ISSUE #1396: DATE and DOUBLE should support more input formats
The [documentation][1] states that for BigQuery data type `DATE`, protocol buffer types `int32` and `int64` are supported while the former is preferred. However, we found that `int64` values are rejected by the Storage Write API because
1. In [BQTableSchemaToProtoDescriptor#49][2], BigQuery data type `DATE` is mapped to protocol buffer type `INT32`.
2. In [JsonToProtoMessage#272][3], Java data type `Integer` is expected for protocol buffer type `INT32` and Java data type `Long` is ignored.

In addition, we could not understand why `DATE` fields do not support date literals, unlike `DATETIME` fields. In [Standard SQL][4], both data types support date literals. You can also think the opposite way: Why does the Storage Write API not expect epoch microseconds for `DATETIME` fields while expecting epoch days for `DATE` fields?

Finally, `JsonToProtoMessage` does not convert `Integer` and `Long` into `Double` for `DOUBLE` fields although the conversion is absolutely safe.

[1]: https://cloud.google.com/bigquery/docs/write-api#data_type_conversions
[2]: https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/BQTableSchemaToProtoDescriptor.java#L49
[3]: https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java#L272
[4]: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#date_type

burakgok:
Fixes #1396

1. Long values are converted into Integer for DATE fields.
2. Date literals are converted into epoch days (Integer) for DATE fields.
3. Number values are converted into Double for DOUBLE fields.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
@@ -25,6 +25,7 @@ import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.Message;
 import com.google.protobuf.UninitializedMessageException;
 import java.math.BigDecimal;
+import java.time.LocalDate;
 import java.util.List;
 import java.util.logging.Logger;
 import org.json.JSONArray;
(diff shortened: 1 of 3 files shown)


yirutang:
Looks great! Could you add unit test for the extra cases supported?
Date with String
Date with Long
Double with Integer

yirutang (approved):
Thanks!

stephaniewang526 (approved):

burakgok:
In #1397, I should not have said "the conversion is absolutely safe" for integer → double conversions since not every integer is accurately representable in IEEE 754. Should I roll back integer → double conversion logic, or will you update test cases? By the way, if [JsonToProtoMessageTest#simpleJsonObjects][1] had a float value, the current logic would fail the test as well.

[1]: https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessageTest.java#L206

PR #1447, opened 2021-12-15
feat: add schema update support in JsonStreamWriter

stephaniewang526:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -80,26 +84,51 @@ private JsonStreamWriter(Builder builder)
...
+      // Close the StreamWriter
+      this.streamWriter.close();
+      // Update JsonStreamWriter's TableSchema and Descriptor
+      this.tableSchema = updatedSchema;
+      this.descriptor =
+          BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
+      // Create a new underlying StreamWriter with the updated TableSchema and Descriptor
+      this.streamWriter =
+          streamWriterBuilder
+              .setUpdatedTableSchema(this.tableSchema)
```
yirutang, line 125: We do we want to set UpdatedTableSchema? I don't think we should provide a setUpdatedTableSchema method on streamWriterBuilder.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -80,26 +84,51 @@ private JsonStreamWriter(Builder builder)
...
+              .setWriterSchema(ProtoSchemaConverter.convert(this.descriptor))
+              .build();
+      // Clear the updateSchema field on the new underlying StreamWriter
+      this.streamWriter.clearUpdateSchema();
```
yirutang, line 129: Since this streamWriter is newly created, we don't need to do this.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -610,6 +627,16 @@ public Builder setTraceId(String traceId) {
       return this;
     }
 
+    /**
+     * Setter for table schema. Used for schema updates.
+     *
+     * @param tableSchema
+     */
+    public Builder setUpdatedTableSchema(TableSchema tableSchema) {
```
yirutang, line 635: Not needed.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -406,6 +408,92 @@ public void testJsonStreamWriterWithDefaultStream()
     }
   }
 
+  @Test
+  public void testJsonStreamWriterSchemaUpdate()
```
yirutang, line 412: We need to add a concurrent append test.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -517,7 +524,15 @@ private AppendRequestAndResponse pollInflightRequestQueue() {
     return new StreamWriter.Builder(streamName);
   }
 
-  /** A builder of {@link StreamWriterV2}s. */
+  public TableSchema getUpdatedSchema() {
```
yirutang, line 527: You need a lock over these get and set to protect concurrent access of the object.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -80,26 +84,51 @@ private JsonStreamWriter(Builder builder)
...
    *
    * @param jsonArr The JSON array that contains JSONObjects to be written
    * @param offset Offset for deduplication
    * @return ApiFuture<AppendRowsResponse> returns an AppendRowsResponse message wrapped in an
    *     ApiFuture
    */
-  public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset) {
+  public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
+      throws IOException, DescriptorValidationException {
+    // Handle schema updates
```
yirutang, line 113: Different appends can be called concurrently, so we need a JsonWriter level lock to make sure only one append is doing these. Add a lock over if (this.streamWriter.getUpdatedSchema() != null) {...}

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
yirutang: I think there should be test already written in that fashion but was removed earlier.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
stephaniewang526: Yes, confirming that `setWriterSchema` with the new `descriptor` is sufficient to update `TableSchema` on the new `StreamWriter`; hence, removing this redundant setter.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
stephaniewang526: Okay. removing the redundant clear method.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
stephaniewang526: Removed this method.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Test case added

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -406,6 +409,208 @@ public void testJsonStreamWriterWithDefaultStream()
...
+      }
+
+      // write rows with updated schema.
+      JSONObject updatedFoo = new JSONObject();
+      updatedFoo.put("col1", "ccc");
+      updatedFoo.put("col2", "ddd");
+      JSONArray updatedJsonArr = new JSONArray();
+      updatedJsonArr.put(updatedFoo);
+      for (int i = 0; i < 10; i++) {
+        ApiFuture<AppendRowsResponse> response3 =
```
yirutang, line 473: Should wait for the response future to come back otherwise test will be unstable.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -406,6 +409,208 @@ public void testJsonStreamWriterWithDefaultStream()
...
+                try {
+                  jsonStreamWriter.append(jsonArr);
+                  next.getAndIncrement();
+                } catch (IOException | DescriptorValidationException e) {
+                  e.printStackTrace();
+                }
+              }
+
+              // update TableSchema
+              Table updatedTable = bigquery.update(updatedTableInfo);
```
yirutang, line 561: I think you should only issue one update request to the table, so move this out and execute it on another thread.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -406,6 +409,208 @@ public void testJsonStreamWriterWithDefaultStream()
...
+              assertEquals(updatedSchema, updatedTable.getDefinition().getSchema());
+
+              // write filler rows bbb until backend acknowledges schema update due to possible
+              // delay
+              for (int w = 0; w < 20; w++) {
+                ApiFuture<AppendRowsResponse> response2 = null;
+                try {
+                  response2 = jsonStreamWriter.append(jsonArr2);
+                  next.getAndIncrement();
+                } catch (IOException | DescriptorValidationException e) {
```
yirutang, line 571: Hopefully with ignore unknown fields, we could keep writing ccc and ddd to the table and expect ddd eventually to show up in the table. This is good for now.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -517,7 +524,12 @@ private AppendRequestAndResponse pollInflightRequestQueue() {
     return new StreamWriter.Builder(streamName);
   }
 
-  /** A builder of {@link StreamWriterV2}s. */
+  /** Thread-safe getter of updated TableSchema */
+  public synchronized TableSchema getUpdatedSchema() {
```
yirutang, line 528: Since it is actually guarded by lock, so just use this.lock.lock() in the method body.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -80,26 +84,50 @@ private JsonStreamWriter(Builder builder)
...
    * @param offset Offset for deduplication
    * @return ApiFuture<AppendRowsResponse> returns an AppendRowsResponse message wrapped in an
    *     ApiFuture
    */
-  public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset) {
+  public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
+      throws IOException, DescriptorValidationException {
+    // Handle schema updates in a Thread-safe way by locking down the operation
+    synchronized (this) {
+      if (this.streamWriter.getUpdatedSchema() != null) {
```
yirutang, line 115: Instead of call getUpdatedSchema twice, could we just call it once?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
yirutang: also increase the thread count of append.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -28,11 +28,14 @@
 import com.google.cloud.bigquery.storage.test.Test.*;
```
yirutang, line 28: I think we should also add a unit test, just the simple case, so that the issue can be discovered by running unit test. In unit test, the update schema can be mimiced by directly return a mock response with table schema.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: sure - updating code to update table in another Thread.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
stephaniewang526: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
stephaniewang526: Done.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Increased Thread count of append from 2 to 5.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
stephaniewang526: Done.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Ack. Will keep this in mind.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
stephaniewang526: Unfortunately, there does not seem to be an elegant way to do that in Java (see [stackoverflow post](https://stackoverflow.com/questions/29537639/check-if-returned-value-is-not-null-and-if-so-assign-it-in-one-line-with-one-m) about this).

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Sync'ed up offline and this is resolved.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Will add the unit test in a separate PR since I'm trying to resolve some intelliJ issue related to `import com.google.cloud.bigquery.storage.test.Test.*;` not being recognized.

yirutang (approved):

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
stephaniewang526: Resolved IDE issue. [Unit test](https://github.com/googleapis/java-bigquerystorage/pull/1447/files#diff-35c9049ac0d019beca1d12aeed7ff3d77812d685e1497daf41c19c6a02b5e619R394) added.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -81,27 +85,49 @@ private JsonStreamWriter(Builder builder)
...
+        // Close the StreamWriter
+        this.streamWriter.close();
+        // Update JsonStreamWriter's TableSchema and Descriptor
+        this.tableSchema = updatedSchema;
+        this.descriptor =
+            BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
+        // Create a new underlying StreamWriter with the updated TableSchema and Descriptor
+        this.streamWriter =
+            streamWriterBuilder
+                .setWriterSchema(ProtoSchemaConverter.convert(this.descriptor))
```
shollyman, line 126: Do you need to propagate anything else to the new writer like stream id, existing trace id, etc or does it all come from properties stored on the JsonStreamWriter?

shollyman (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
stephaniewang526: Yes, those things are needed and are already being populated in the `streamWriterBuilder`.

PR #1455, opened 2021-12-28
feat: Add ignoreUnknownField support in JsonWriter

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -137,7 +137,8 @@ public class JsonStreamWriter implements AutoCloseable {
     for (int i = 0; i < jsonArr.length(); i++) {
       JSONObject json = jsonArr.getJSONObject(i);
       Message protoMessage =
-          JsonToProtoMessage.convertJsonToProtoMessage(this.descriptor, this.tableSchema, json, ignoreUnknownFields);
+          JsonToProtoMessage.convertJsonToProtoMessage(
+              this.descriptor, this.tableSchema, json, ignoreUnknownFields);
       rowsBuilder.addSerializedRows(protoMessage.toByteString());
     }
     // Need to make sure refreshAppendAndSetDescriptor finish first before this can run
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -355,12 +355,14 @@ public class JsonStreamWriter implements AutoCloseable {
     }
 
     /**
-     * Ignores fields unknown to BigQuery in Json input.
+     * Setter for a ignoreUnkownFields, if true, unknown Json fields to BigQuery will be ignored
+     * instead of error out.
      *
+     * @param ignoreUnknownFields
      * @return Builder
      */
-    public Builder ignoreUnknownFields() {
-      this.ignoreUnknownFields = true;
+    public Builder setIgnoreUnknownFields(boolean ignoreUnknownFields) {
+      this.ignoreUnknownFields = ignoreUnknownFields;
       return this;
     }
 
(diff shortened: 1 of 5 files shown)


stephaniewang526:
We should revisit the [test case for schema update](https://github.com/googleapis/java-bigquerystorage/pull/1447/files/2ba2b894e2bc9d48e4fc777db5e831c859f7277c#diff-ffee812085dc0a657a9349957229bd7aaa7a4cd150027e4435faf3eed5023b6bR497) while introducing this feat (related to [this comment](https://github.com/googleapis/java-bigquerystorage/pull/1447#discussion_r774182728)).

yirutang (commented):
PTAL, rebased.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -350,6 +354,16 @@ public Builder setTraceId(String traceId) {
       return this;
     }
 
+    /**
+     * Ignores fields unknown to BigQuery in Json input.
+     *
+     * @return Builder
+     */
+    public Builder ignoreUnknownFields() {
```
stephaniewang526, line 362: I think it's more conventional and explicit(true vs. false) if we do `setIgnoreUnknownFields(boolean ignoreUnknwonFields) {...}` ([an equivalent example](https://github.com/googleapis/java-bigquery/blob/main/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/QueryJobConfiguration.java#L497-L499) in google-cloug-bigquery client).

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -64,7 +64,8 @@ public static DynamicMessage convertJsonToProtoMessage(Descriptor protoSchema, J
     Preconditions.checkNotNull(protoSchema, "Protobuf descriptor is null.");
     Preconditions.checkState(json.length() != 0, "JSONObject is empty.");
 
-    return convertJsonToProtoMessageImpl(protoSchema, null, json, "root", /*topLevel=*/ true);
+    return convertJsonToProtoMessageImpl(
+        protoSchema, null, json, "root", /*topLevel=*/ true, false);
```
stephaniewang526, line 68: As discussed offline, `topLevel` is not used in `convertJsonToProtoMessageImpl`. Hence, we should remove it.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -117,9 +151,11 @@ private static DynamicMessage convertJsonToProtoMessageImpl(
       String jsonLowercaseName = jsonName.toLowerCase();
       String currentScope = jsonScope + "." + jsonName;
       FieldDescriptor field = protoSchema.findFieldByName(jsonLowercaseName);
-      if (field == null) {
+      if (field == null && !ignoreUnknownFields) {
         throw new IllegalArgumentException(
             String.format("JSONObject has fields unknown to BigQuery: %s.", currentScope));
+      } else if (field == null) {
```
stephaniewang526, line 157: This check seems redundant.

stephaniewang526 (changes requested):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
yirutang: Ack

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
yirutang: If it is null, it means we cannot proceed on further processing. Otherwise, null ptr will be thrown if we continue.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
stephaniewang526: Okay makes sense -- this allows us to skip the unknown field in the middle and proceed onto other fields.

stephaniewang526 (approved):

PR #1456, opened 2021-12-30
fix: update storageError support due to server side enhancement

stephaniewang526:
client-side changes corresponding to cl/408735437

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
@@ -594,14 +594,11 @@ public class StreamWriter implements AutoCloseable {
                   + " for stream "
                   + streamName);
         } else {
-          this.connectionFinalStatus = finalStatus;
+          Exceptions.StorageException storageException = Exceptions.toStorageException(finalStatus);
+          this.connectionFinalStatus = storageException != null ? storageException : finalStatus;
           log.info(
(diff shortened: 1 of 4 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -599,13 +599,13 @@ private void doneCallback(Throwable finalStatus) {
...
         }
       }
+      Exceptions.StorageException storageException = Exceptions.toStorageException(finalStatus);
+      if (storageException != null) {
+        this.connectionFinalStatus = storageException;
+      }
     } finally {
       this.lock.unlock();
     }
-    Exceptions.StorageException storageException = Exceptions.toStorageException(finalStatus);
```
yirutang, line 605: why move this inside the lock?

yirutang: Actually, we should move it even up to l597, otherwise, there will be a problem.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
stephaniewang526: As discussed offline, the Exception conversion needs to be protected by the lock so that it is not thrown before getting converted.

tswast (approved):

PR #1470, opened 2022-01-07
ci: try using oidc for samples test

chingor13:


.github/workflows/samples.yaml
```diff
@@ -12,3 +16,65 @@ jobs:
...
+    - run: java -version
+    - run: echo "${SAMPLE_SECRETS}" > ${{ runner.temp }}/sample-secrets
+      env:
+        SAMPLE_SECRETS: '${{ steps.secrets.outputs.env_vars }}'
+    - run: .kokoro/build.sh
+      env:
+        JOB_TYPE: samples
+        GOOGLE_CLOUD_PROJECT: java-docs-samples-testing
+        SAMPLE_SECRETS: ${{ runner.temp }}/sample-secrets
+        KOKORO_GITHUB_PULL_REQUEST_NUMBER: 123
```
chingor13, line 58: We can clean up this and the build.sh to signal to only run the snapshot test

.github/workflows/samples.yaml
```diff
@@ -12,3 +16,65 @@ jobs:
...
+        service_account: 'java-docs-samples-testing@java-docs-samples-testing.iam.gserviceaccount.com'
+        create_credentials_file: 'true'
+        access_token_lifetime: 300s
+        token_format: 'access_token'
+        cleanup_credentials: false
+    - id: 'secrets'
+      uses: 'google-github-actions/get-secretmanager-secrets@v0'
+      with:
+        secrets: |-
+          env_vars:cloud-devrel-kokoro-resources/java-bigquerystorage-samples-secrets
```
chingor13, line 44: With this strategy, we can also put this secret in the java-docs-samples-testing project instead of the shared project

PR #1550, opened 2022-02-25
fix: add more idiomatic way to insert ARRAY<BYTES> data

stephaniewang526:
Towards b/219375453

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigquery.storage.v1;
 
 import com.google.api.pathtemplate.ValidationException;
+import com.google.cloud.ByteArray;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -416,7 +416,9 @@ private static void fillRepeatedField(
             }
           }
           if (!added) {
-            if (val instanceof JSONArray) {
+            if (val instanceof byte[]) {
+              protoMsg.addRepeatedField(fieldDescriptor, val);
+            } else if (val instanceof JSONArray) {
```
tswast, line 421: [no action required in this PR] It's a bummer to still see `org.json` used as a dependency here. I assume we have an issue tracking a migration to Gson?

tswast (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
stephaniewang526: Yeah, for sure -- b/216657461 is tracking this work.

PR #1563, opened 2022-03-04
feat: add support to a few more specific StorageErrors for the Write API

stephaniewang526:
OFFSET_OUT_OF_RANGE
OFFSET_ALREADY_EXISTS
STREAM_NOT_FOUND

Towards b/220198094

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -78,6 +78,27 @@ public final class Exceptions {
     }
   }
 
+  /** Offset already exists. */
+  public static final class OffsetAlreadyExists extends StorageException {
+    protected OffsetAlreadyExists(Status grpcStatus, Metadata metadata, String name) {
+      super(grpcStatus, metadata, name, ImmutableMap.of());
+    }
+  }
+
+  /** Offset out of range. */
+  public static final class OffsetOutOfRange extends StorageException {
+    protected OffsetOutOfRange(Status grpcStatus, Metadata metadata, String name) {
+      super(grpcStatus, metadata, name, ImmutableMap.of());
+    }
+  }
+
+  /** Stream is not found. */
+  public static final class StreamNotFound extends StorageException {
+    protected StreamNotFound(Status grpcStatus, Metadata metadata, String name) {
+      super(grpcStatus, metadata, name, ImmutableMap.of());
+    }
+  }
+
   private static StorageError toStorageError(com.google.rpc.Status rpcStatus) {
     for (Any detail : rpcStatus.getDetailsList()) {
       if (detail.is(StorageError.class)) {
@@ -112,6 +133,15 @@ public final class Exceptions {
 
       case SCHEMA_MISMATCH_EXTRA_FIELDS:
         return new SchemaMismatchedException(grpcStatus, null, error.getEntity());
+
+      case OFFSET_OUT_OF_RANGE:
+        return new OffsetOutOfRange(grpcStatus, null, error.getEntity());
+
+      case OFFSET_ALREADY_EXISTS:
+        return new OffsetAlreadyExists(grpcStatus, null, error.getEntity());
+
+      case STREAM_NOT_FOUND:
+        return new StreamNotFound(grpcStatus, null, error.getEntity());
 
       default:
         return null;


yirutang (commented):
Add a e2e test for all these exceptions + to show the StatusRuntimeException still works?

stephaniewang526:
Converting this to draft as we realized that there are backend changes needed (https://b.corp.google.com/issues/220198094#comment7) before proceeding.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -78,6 +78,27 @@ protected SchemaMismatchedException(Status grpcStatus, Metadata metadata, String
     }
   }
 
+  /** Offset already exists. */
+  public static final class OffsetAlreadyExists extends StorageException {
```
yirutang, line 82: could we make this exception richer by providing the actual and expected offset? It can be done by parsing the error message.

yirutang:
The exception backend change is fully rolled out.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -106,12 +141,32 @@ public static StorageException toStorageException(
     if (error == null) {
       return null;
     }
+    String streamName = error.getEntity();
+    String errorMessage =
```
yirutang, line 145: Add a comment about this is to trim the entity sub message. Also narrow it further down by looking for "Entity: ".

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -106,12 +141,32 @@ public static StorageException toStorageException(
...
+        error.getErrorMessage().indexOf("Entity") > 0
+            ? error.getErrorMessage().substring(0, error.getErrorMessage().indexOf("Entity")).trim()
+            : error.getErrorMessage().trim();
+    // Note: if errorMessage format changes, the parsing logic below might break
+    Long expectedOffet =
+        Long.parseLong(
+            errorMessage.substring(
+                errorMessage.lastIndexOf("offset") + 7, errorMessage.lastIndexOf(",")));
```
yirutang, line 153: Have a whole stream match pattern for "expected offset $0, received $1". I am writing cl/445470232 to make sure the parsing will not break.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: I cannot just look for "Entity:" as of now since it is somehow missing from the detailMessage of `OFFSET_ALREADY_EXISTS` error. The full detailMessage I get looks like: "ALREADY_EXISTS: The offset is within stream, expected offset 1, received 0" (instead of "ALREADY_EXISTS: The offset is within stream, expected offset 1, received 0 Entity: projects/.../datasets/.../tables/testtable/streams/CiQ2NGY3OT...").

yirutang: So why do you need to trim the next message?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
stephaniewang526: What does "next message" mean? I need to trim the message of whitespaces either way (with or without Entity) in order to get the "expected offset [0-9]+, received [0-9]+" portion to parse out the offset numbers.

yirutang (approved):

PR #1603, opened 2022-03-31
feat: Rewrite Write API samples to match best practices

gnanda:
- split samples into initialization, append, cleanup
- Change appends to async


samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -26,14 +29,21 @@
...
 import com.google.protobuf.Descriptors.DescriptorValidationException;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Phaser;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
 public class WriteToDefaultStream {
 
+  private final Phaser inflightRequestPhaser = new Phaser(1);
```
stephaniewang526, line 42: We don't prefer having anything out here like this, they should go inside either the driver method (`runWriteToDefaultStream`) or the body of the sample (`writeToDefaultStream`). Also, we need comment (explanation) around why we are using `inflightRequestPhaser`.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -66,15 +69,73 @@ public static void writeToDefaultStream(String projectId, String datasetName, St
...
+    // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.html
+    writer.streamWriter = JsonStreamWriter.newBuilder(parentTable.toString(), tableSchema).build();
+    return writer;
+  }
+
+  public void append(JSONArray jsonArr)
+      throws DescriptorValidationException, IOException, ExecutionException {
+    ApiFuture<AppendRowsResponse> future = this.streamWriter.append(jsonArr);
+    this.inflightRequestPhaser.register();
+    ApiFutures.addCallback(
```
stephaniewang526, line 106: Can we add a comment here to explain this?

stephaniewang526 (changes requested):

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
```diff
@@ -17,22 +17,34 @@
...
 import com.google.cloud.bigquery.storage.v1.WriteStream;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.Descriptors.DescriptorValidationException;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Phaser;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
 public class WriteCommittedStream {
```
yirutang, line 39: This name is a little confusing, maybe name it to WriteCommittedStreamSample? I am not sure the impact of changing the sample's name.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
```diff
@@ -45,54 +57,104 @@ public static void runWriteCommittedStream()
...
+    // To disable deduplication, omit the offset or use WriteStream.Type.DEFAULT.
+    ApiFuture<AppendRowsResponse> future = streamWriter.append(data, /*offset=*/ offset);
+    this.inflightRequestPhaser.register();
+    ApiFutures.addCallback(
+        future, new AppendCompleteCallback(this), MoreExecutors.directExecutor());
+  }
+
+  public void cleanup() {
+    // Wait for all in-flight requests to complete.
+    this.inflightRequestPhaser.arriveAndAwaitAdvance();
```
yirutang, line 127: Close the JsonWriter here? Since you are doing a Finalize.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
VeronicaWasson: I think most of our samples don't include "Sample" in the name. e.g., [createTable](https://github.com/googleapis/java-bigquery/blob/0f19738e5e4229a52c8cc6b40d3bace40f8df747/samples/snippets/src/main/java/com/example/bigquery/CreateTable.java#L44).

Maybe WriteToCommittedStream?

yirutang: Make sense. Currently it sounded like a Stream. Will we just create a new file, switch the pointer and delete the old file?

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
```diff
@@ -45,54 +57,104 @@ public static void runWriteCommittedStream()
 
   public static void writeCommittedStream(String projectId, String datasetName, String tableName)
       throws DescriptorValidationException, InterruptedException, IOException {
+    // One time initialization.
```
VeronicaWasson, line 60: Maybe put the initalization code into a separate writer.initialize() method? I really like the structure of https://github.com/googleapis/java-bigquerystorage/pull/1585 from a docs perspective.

samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
```diff
@@ -48,63 +60,124 @@ public static void runWritePendingStream()
...
+
+    AppendCompleteCallback(WritePendingStream parent) {
+      this.parent = parent;
+    }
+
+    public void onSuccess(AppendRowsResponse response) {
+      if (response.hasError()) {
+        System.out.format("Error: %s\n", response.getError());
+      } else {
+        System.out.format("Append %d success\n", response.getAppendResult().getOffset().getValue());
```
yirutang, line 169: For pending stream, show user a way to set stream_status_ and append to check on the status to fail the append?

samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
```diff
@@ -48,63 +60,124 @@ public static void runWritePendingStream()
...
+      System.out.println("Failed to append records. \n" + e);
+    }
+
+    // Final cleanup for the stream.
+    writer.cleanup();
+    System.out.println("Appended records successfully.");
+
+    // Once all streams are done, commit all of them in one request. This example only has the one
+    // stream.
+    BatchCommitWriteStreamsRequest commitRequest =
```
yirutang, line 94: We shouldn't commit for failed writes?

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -66,15 +69,73 @@ public static void writeToDefaultStream(String projectId, String datasetName, St
...
+
+    private final WriteToDefaultStream parent;
+
+    AppendCompleteCallback(WriteToDefaultStream parent) {
+      this.parent = parent;
+    }
+
+    public void onSuccess(AppendRowsResponse response) {
+      if (response.hasError()) {
+        System.out.format("Error: %s\n", response.getError());
```
yirutang, line 125: show a simple retry here?

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
```diff
@@ -45,54 +57,104 @@ public static void runWriteCommittedStream()
...
+  class AppendCompleteCallback implements ApiFutureCallback<AppendRowsResponse> {
+
+    private final WriteCommittedStream parent;
+
+    AppendCompleteCallback(WriteCommittedStream parent) {
+      this.parent = parent;
+    }
+
+    public void onSuccess(AppendRowsResponse response) {
+      if (response.hasError()) {
```
yirutang, line 143: I don't know what to do here...Maybe add a comment of how this can be retried? If we don't have a good retry case, should we have this sample?

VeronicaWasson: My $0.02: I think it's better to show how to check the AppendRowsResponse error status, rather than not show anything here.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -45,19 +55,12 @@ public static void runWriteToDefaultStream()
...
+    // One time initialization.
+    WriteToDefaultStream writer = WriteToDefaultStream.create(projectId, datasetName, tableName);
 
-    // Use the JSON stream writer to send records in JSON format.
-    // For more information about JsonStreamWriter, see:
-    // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.html
-    try (JsonStreamWriter writer =
-        JsonStreamWriter.newBuilder(parentTable.toString(), tableSchema).build()) {
+    try {
       // Write two batches to the stream, each with 10 JSON records. A writer should be used for as
```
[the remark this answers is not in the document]
gnanda: Added a comment.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -66,15 +69,73 @@ public static void writeToDefaultStream(String projectId, String datasetName, St
...
+    }
+
+    // Final cleanup.
+    writer.cleanup();
+    System.out.println("Appended records successfully.");
+  }
+
+  public static WriteToDefaultStream create(String projectId, String datasetName, String tableName)
+      throws DescriptorValidationException, IOException, InterruptedException {
+    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();
```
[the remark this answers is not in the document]
gnanda: Will do in followup; I changed this to just be PendingStream which encompasses the other cases, so we can finalize on a design in one place and replicate to the others.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
gnanda: Added a comment about the Phaser (which we need in order to wait for all responses before closing the stream)

Discussed the variables offline: the Write API is a special case where we need a stateful client and want to make explicit that users should keep the writer around for as long as possible.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
gnanda: Added

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
gnanda: sgtm  I'll do this in a followup

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
gnanda: Added the close, but as we discussed offline, we still need to wait for inflight requests to finish

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
gnanda: Changed it back. Not sure how to do it in a way that both exemplifies the structure and is "Good Java"

samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
gnanda: Done, thought I'm not great at Java, so maybe there's a better way?

samples/snippets/src/main/java/com/example/bigquerystorage/WritePendingStream.java
gnanda: Added an exception throw in cleanup and a comment about only committing successful streams.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteCommittedStream.java
gnanda: IIUC StreamWriter already handles this field which will invoke onFailure instead 
https://github.com/googleapis/java-bigquerystorage/blob/54874be945c2b88be6be03ae654277445c17741d/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L565

I updated that code to show a little error handling; not sure if there's something more specific to show there

gnanda:
Updated with comments and only changing WritePendingStream for now since it encompasses the other cases.  Once we finalize on a plan, I'll update the other 2 to match the same structure

yirutang (approved):

stephaniewang526 (approved):

PR #1637, opened 2022-04-29
fix: A stuck when the client fail to get DoneCallback

yirutang:
Add a timeout of one minute waiting for done callback to be called. Same timeout as client close.
The donecallback mainly gives back the server side error status, so it is not critical. In Dataflow connector, we saw hang because the DoneCallback is lost and we wait forever on it.

Stack trace in b/230501926

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
@@ -506,9 +506,9 @@ public class StreamWriter implements AutoCloseable {
       }
       Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
     }
-    connectionFinalStatus = new StatusRuntimeException(
-        Status.fromCode(Code.CANCELLED)
-            .withDescription("Timeout waiting for DoneCallback."));
+    connectionFinalStatus =
+        new StatusRuntimeException(
+            Status.fromCode(Code.CANCELLED).withDescription("Timeout waiting for DoneCallback."));
     return;
   }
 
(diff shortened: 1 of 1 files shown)


yirutang:
@reuvenlax

reuvenlax:
LGTM

gnanda:
Do we know why the doneCallback isn't being called? Is there any chance of a memory leak from this because resources aren't being cleaned up?

e.g. do we need a catch block here to set connectionFinalStatus in case of an exception throw?
https://github.com/googleapis/java-bigquerystorage/blob/3a33f49e2796f25dc271596ea032094e6cd72d23/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L622

yirutang:
> Do we know why the doneCallback isn't being called? Is there any chance of a memory leak from this because resources aren't being cleaned up?
> 
> Otherwise LGTM

I don't exactly know, what we saw is the connector stuck for days in this: https://b.corp.google.com/issues/230501926#comment3

According to Reuven:
Reuven Lax, Wed 11:34 AM
Slava says that they've seen cases where the done callback doesn't come after closing the stream

Reuven Lax, Wed 11:37 AM
he also says that we had a lot of trouble with very long-lived streaming rpcs. what we ended up switching to was setting a 3 minute timeout on the streaming rpc, and reconnecting every 3 minutes

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -505,6 +506,14 @@ private void waitForDoneCallback() {
       }
       Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
     }
+    this.lock.lock();
+    if (connectionFinalStatus == null) {
```
gnanda, line 510: I'm not super familiar w/java, but do we need the try {} finallly {lock.unlock();} around this block like I see in other places, or is that unnecessary?

gnanda (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
yirutang: Sure.

stephaniewang526 (approved):

PR #1638, opened 2022-05-03
fix: Fix a possible NULL PTR after introduced timeout on waitForDone

yirutang:
This PR:
https://github.com/googleapis/java-bigquerystorage/pull/1637

introduced a timeout for waitForDone. So we could potentially drained the inflight requests when a RequestCallback comes back. Add process for null when peak the inflight queue.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
@@ -573,6 +573,9 @@ public class StreamWriter implements AutoCloseable {
         conectionRetryCountWithoutCallback = 0;
       }
       requestWrapper = pollInflightRequestQueue();
+      if (requestWrapper == null) {
+        return;
+      }
     } finally {
       this.lock.unlock();
     }
@@ -640,11 +643,17 @@ public class StreamWriter implements AutoCloseable {
 
   @GuardedBy("lock")
   private AppendRequestAndResponse pollInflightRequestQueue() {
-    AppendRequestAndResponse requestWrapper = this.inflightRequestQueue.pollFirst();
-    --this.inflightRequests;
-    this.inflightBytes -= requestWrapper.messageSize;
-    this.inflightReduced.signal();
-    return requestWrapper;
+    if (!this.inflightRequestQueue.isEmpty()) {
+      AppendRequestAndResponse requestWrapper = this.inflightRequestQueue.pollFirst();
+      --this.inflightRequests;
+      this.inflightBytes -= requestWrapper.messageSize;
+      this.inflightReduced.signal();
+      return requestWrapper;
+    } else {
+      // It is possible when requestCallback is called, the inflight queue is already drained to do timeout waiting
+      // for done.
+      return null;
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -573,6 +573,9 @@ private void requestCallback(AppendRowsResponse response) {
         conectionRetryCountWithoutCallback = 0;
       }
       requestWrapper = pollInflightRequestQueue();
+      if (requestWrapper == null) {
```
gnanda, line 576: If we haven't cleaned up and this happens, we should throw an exception

Maybe instead of using null, check if we've cleaned up, and don't call the method (since this is under lock anyways)?

yirutang:
@gaolemeng

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
yirutang: If we have cleaned up, then all inflight requests are all set with an exception already. 
I don't think we should throw on requestCallback?

gnanda: I meant the opposite: if we haven't cleaned up.  

If during normal operations we get a callback but there's nothing in the queue, that should never happen, so we should throw (or somehow error out)

yirutang: Updated the code, hope it is more straightforward.

gnanda: Yeah it is, but I was thinking we'd also have

onDone {
  this.done = true;
}

and in your change:
if (!this.done && this.inflightRequestQueue.isEmpty()) {
  // Handle this error case!
}

But since it's in the callback, that's probably more complicated than I originally thought... Should we handle that in a followup?

yirutang: PTAL

gnanda (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
yirutang: There is no super good place for done, since the lock is done inside of the cleanupInflightQueue.

yirutang: Added an error message, without a response, user will probably not be able to get the actual exception.

stephaniewang526:
@yirutang 
There is a lint issue on this PR: can be fixed by running `mvn com.coveo:fmt-maven-plugin:format`
And OwlBot is pending: can be fixed by tagging this PR `owlbot: run`
Then you should be able to merge it in (without my approval).

PR #1642, opened 2022-05-06
fix: Add a throwException behavior when the StreamWriter inflight queue is full

### LINKED ISSUE #1539: BigQuery Storage Write API: LimitExceededBehavior of FlowControlSettings is ignored in JsonStreamWriter
#### Environment details
Java version: 8 latest
Library version(s): 2.8.1

#### Steps to reproduce
  1. Create a new JsonStreamWriter with FlowControlSettings set
  ```
FlowControlSettings flowControlSettings = FlowControlSettings.newBuilder()
        .setLimitExceededBehavior(FlowController.LimitExceededBehavior.ThrowException)
        .build();
JsonStreamWriter writer = JsonStreamWriter.newBuilder(tableName, tableSchema)
        .setFlowControlSettings(flowControlSettings)
        .build();
```
  2. Simulate service unavailability. After reaching the inflight quota limit `writer.append(jsonArr)` blocks indefinitely.

#### Stack trace
```
io.vertx.core.VertxException: Thread blocked at sun.misc.Unsafe.park(Native Method) at
    java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215) at
    java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2163) at
    com.google.cloud.bigquery.storage.v1.StreamWriter.maybeWaitForInflightQuota(StreamWriter.java:322) at
    com.google.cloud.bigquery.storage.v1.StreamWriter.appendInternal(StreamWriter.java:310) at
    com.google.cloud.bigquery.storage.v1.StreamWriter.append(StreamWriter.java:272) at
    com.google.cloud.bigquery.storage.v1.JsonStreamWriter.append(JsonStreamWriter.java:147) at
    com.google.cloud.bigquery.storage.v1.JsonStreamWriter.append(JsonStreamWriter.java:99)
    ...
```

During service problems like omg/48020 or other changes, that affect writing data to BigQuery we must avoid to indefinetly get blocked in our application. We can't afford that our application crashes or hangs because we are not able to ingest data to BigQuery. Instead we would rather have our application discarding those BigQuery events and trigger an alert. If the library would throw an exception instead of blocking, which this LimitExceededBehaviour seems to be intended for, this would be sufficient for us. So in this case WE could decide what to do when the limit is reached and not the library decides to block. This has real dramatic impact if this is permanently blocking due to a bug like omg/48020 and never recovers. To work around this with future bugs we implemented the same inflight quota counting on top and start discarding events before the the library internal quotas are reached. This is obviously not a good solution.

### PR DESCRIPTION
**Pull request by yirutang**
Add a throwException behavior when the StreamWriter inflight queue is full.

Fixes #1539 ☕️

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigquery.storage.v1;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.SettableApiFuture;
+import com.google.api.gax.batching.FlowController;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.rpc.FixedHeaderProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
@@ -73,6 +74,11 @@ public class StreamWriter implements AutoCloseable {
    * Max allowed inflight bytes in the stream. Method append is blocked at this.
    */
   private final long maxInflightBytes;
+
+  /*
+   * Behavior when inflight queue is exceeded. Only supports Block or Throw, default is Block.
+   */
+  private final FlowController.LimitExceededBehavior limitExceededBehavior;
 
   /*
    * TraceId for debugging purpose.
@@ -190,6 +196,7 @@ public class StreamWriter implements AutoCloseable {
     this.writerSchema = builder.writerSchema;
     this.maxInflightRequests = builder.maxInflightRequest;
     this.maxInflightBytes = builder.maxInflightBytes;
+    this.limitExceededBehavior = builder.limitExceededBehavior;
     this.traceId = builder.traceId;
     this.waitingRequestQueue = new LinkedList<AppendRequestAndResponse>();
     this.inflightRequestQueue = new LinkedList<AppendRequestAndResponse>();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by GaoleMeng — APPROVED** [review]

File: google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java [hunk 1]
```diff
@@ -332,18 +339,29 @@ private void maybeWaitForInflightQuota() {
...
-            "Interrupted while waiting for inflight quota. Stream: "
-                + streamName
-                + " Error: "
-                + e.toString());
+      if (this.limitExceededBehavior == FlowController.LimitExceededBehavior.ThrowException) {
+        throw new StatusRuntimeException(
+            Status.fromCode(Code.RESOURCE_EXHAUSTED)
+                .withDescription(
+                    "Exceeds client side inflight buffer, consider add more buffer or open more connections."));
+      } else if (this.limitExceededBehavior == FlowController.LimitExceededBehavior.Ignore) {
```
tkosse, line 347: Wouldn't be the ignore option just a break; at this point?

yirutang: L735, by default it is Block.

yirutang: https://github.com/googleapis/java-bigquerystorage/pull/1643

tkosse: What I meant was, that the actual meaning of LimitExceededBehavior.Ignore is, that this inflight quota counting is simply ignored. So the StreamWriter just continues sending. But this option is probably not that relevant for productive setups.
```suggestion
      } else if (this.limitExceededBehavior == FlowController.LimitExceededBehavior.Ignore) {
        break;
      }
```

yirutang: I tend to think we shouldn't allow it...

PR #1643, opened 2022-05-06
fix: add extra JsonWriterTest to show that the LimitBehavior addition is not breaking

yirutang:
Also change the exception rejection into before the request was added to the queue instead of after.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -78,10 +78,6 @@ public class JsonStreamWriter implements AutoCloseable {
     this.protoSchema = ProtoSchemaConverter.convert(this.descriptor);
     this.totalMessageSize = protoSchema.getSerializedSize();
     streamWriterBuilder.setWriterSchema(protoSchema);
-    if (builder.flowControlSettings != null) {
-      streamWriterBuilder.setLimitExceededBehavior(
-          builder.flowControlSettings.getLimitExceededBehavior());
-    }
     setStreamWriterSettings(
         builder.channelProvider,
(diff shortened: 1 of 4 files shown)


yirutang:
@tkosse

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -78,10 +78,6 @@ private JsonStreamWriter(Builder builder)
     this.protoSchema = ProtoSchemaConverter.convert(this.descriptor);
     this.totalMessageSize = protoSchema.getSerializedSize();
     streamWriterBuilder.setWriterSchema(protoSchema);
-    if (builder.flowControlSettings != null) {
```
yirutang, line 81: It is set inside setStreamWriterSettings, remove duplicate code.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriterTest.java
```diff
@@ -581,4 +582,25 @@ public void run() throws Throwable {
                   "Exceeds client side inflight buffer, consider add more buffer or open more connections"));
     }
   }
+
+  @Test
+  public void testFlowControlSettingNoLimitBehavior() throws Exception {
```
GaoleMeng, line 588: Discussed offline, from this test and the previous test I'm thinking the current behavior is not so ideal, in the throw exception situation, since we may successfully send out the request, however we will still throw exception
the client might think they need to resend the request, however they don't need.

yirutang: good catch, d.

stephaniewang526 (approved):

PR #1648, opened 2022-05-07
fix: doubles and floats as numerics

### LINKED ISSUE #1516: BigQuery StorageWrite: JsonToProtoMessage numeric error
#### Environment details

1. BigQuery StorageWrite
2. OS type and version: MacOS 11.6.3
3. Java version: 11
4.  version(s): 2.8.4

#### Steps to reproduce

  1. Create a bigquery table with only one field of type NUMERIC and with name "d"
  2. Use JsonStreamWriter with this JSON Array `[{"d":1}]`
  3. run

#### Stack trace
```
java.lang.IllegalArgumentException: JSONObject does not have a bytes field at root.d.
	at com.google.cloud.bigquery.storage.v1.JsonToProtoMessage.fillField(JsonToProtoMessage.java:351)
	at com.google.cloud.bigquery.storage.v1.JsonToProtoMessage.convertJsonToProtoMessageImpl(JsonToProtoMessage.java:176)
	at com.google.cloud.bigquery.storage.v1.JsonToProtoMessage.convertJsonToProtoMessage(JsonToProtoMessage.java:115)
	at com.google.cloud.bigquery.storage.v1.JsonStreamWriter.append(JsonStreamWriter.java:140)
	at com.google.cloud.bigquery.storage.v1.JsonStreamWriter.append(JsonStreamWriter.java:99)
	at persistence.BigqueryStreamingClient.streamJsonArray(BigqueryStreamingClient.java:80)
```

#### External references such as API reference guides

- ?

It seems if you try to write an int/long to a NUMERIC field, it fails because the proto converter requires a bytes object for NUMERIC's. Looking at https://github.com/googleapis/java-bigquerystorage/blob/4c532b0ff9b7dba0105d24accc71e0714d488dbb/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java#L229 it looks like only Strings or ByteStrings are converted to bytes. The end result is not being able to write numeric java types to the NUMERIC BQ column.

dark0dave:
Allows floating point numbers to be numerics and big numerics.
Fixes #1516

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
@@ -296,6 +296,12 @@ public class JsonToProtoMessage {
                   BigDecimalByteStringEncoder.encodeToNumericByteString(
                       new BigDecimal(((Number) val).longValue())));
               return;
+            } else if (val instanceof Float || val instanceof Double) {
+              protoMsg.setField(
+                  fieldDescriptor,
+                  BigDecimalByteStringEncoder.encodeToNumericByteString(
+                      new BigDecimal(((Number) val).doubleValue())));
+              return;
             }
           } else if (fieldSchema.getType() == TableFieldSchema.Type.BIGNUMERIC) {
             if (val instanceof String) {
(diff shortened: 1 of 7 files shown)


stephaniewang526:
Hi, any update on this?

dark0dave:
yep fixing up now.

dark0dave:
let me fix the tests

dark0dave:
@stephaniewang526 appologies, been busy. Going to run tests now to see if I missed anything.

dark0dave:
@stephaniewang526 should be good now.

stephaniewang526 (approved):

feamcor:
This code `new BigDecimal(((Number)val).doubleValue()` is problematic as it generates an instance of `BigDecimal` (that is historically flawed) with a scale greater than what `NUMERIC` and `BIGNUMERIC` can cope with (9 and 38 digits specifically).

Example: A `Double` of value `133.36` is converted to a `BigDecimal` of value `133.3600000000000136424205265939235687255859375`, resulting on the following exception: `java.lang.IllegalArgumentException: ByteString scale cannot exceed 9: 133.3600000000000136424205265939235687255859375`.

I recommend to change the code to take the `Float`/`Double` convert it to `String` and create the `BigDecimal` from the `String`.

Thanks.

PR #1682, opened 2022-06-22
feat: increase timeout waiting for done callback from 2 minutes to 5 …

yirutang:
…minutes

In my stress test I am seeing a lot of CANCELLED: Timeout waiting for DoneCallback. failure when I am making calls to EU region, which led me to think the timeout is too short.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,20 +49,20 @@ If you are using Maven without BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:25.2.0')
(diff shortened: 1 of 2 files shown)


prash-mi:
@yirutang Thanks for raising the PR. Quick question - where did you run the code from (local vs a VM?), also did the issue resolve with 5 minutes as the timeout?

alicejli:
@yirutang gentle bump. Could you elaborate on your findings from your stress test? Thanks!

yirutang:
> @yirutang Thanks for raising the PR. Quick question - where did you run the code from (local vs a VM?), also did the issue resolve with 5 minutes as the timeout?

I ran it from a VM.

PR #1726, opened 2022-08-01
feat!: remove a dummy WriterClosedException

yirutang:
Remove a dummy WriterClosedException. Because we now have StreamWriterClosedException, remove the WriterClosedException to avoid confusion.

The WriterClosedException is used as a place holder when we cleanup Inflight request. In call cases, we should have gotten a not null connectionFinalStatus in WaitForDoneCallback:
https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L533

So no user should be able to get this exception.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,13 @@ implementation 'com.google.cloud:google-cloud-bigquerystorage'
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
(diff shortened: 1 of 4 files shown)


shollyman (approved):
I think you can revert the README changes, otherwise LGTM.

PR #1752, opened 2022-08-16
fix: Add missing field error to row error message

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -239,6 +239,35 @@ public final class Exceptions {
     }
   }
 
+  /** This exception is used internally to handle field level parsing errors. */
+  public static class FieldParseError extends IllegalArgumentException {
+    private final String fieldName;
+    private final String bqType;
+    private final Throwable cause;
+
+    public FieldParseError(String fieldName, String bqType, Throwable cause) {
+      this.fieldName = fieldName;
+      this.bqType = bqType;
+      this.cause = cause;
+    }
+
+    public String getFieldName() {
+      return fieldName;
+    }
+
+    public String getBqType() {
+      return bqType;
+    }
+
+    public Throwable getCause() {
+      return cause;
+    }
+
+    public String getMessage() {
+      return cause.getMessage();
+    }
+  }
+
   /**
    * This writer instance has either been closed by the user explicitly, or has encountered
    * non-retriable errors.
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -237,7 +237,7 @@ public final class Exceptions {
     private final String streamName;
 
     public AppendSerializtionError(String streamName, Map<Integer, String> rowIndexToErrorMessage) {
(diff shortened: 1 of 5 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -221,11 +221,22 @@ private static DynamicMessage convertJsonToProtoMessageImpl(
...
+      try {
+        if (!field.isRepeated()) {
+          fillField(
+              protoMsg, field, fieldSchema, json, jsonName, currentScope, ignoreUnknownFields);
+        } else {
+          fillRepeatedField(
+              protoMsg, field, fieldSchema, json, jsonName, currentScope, ignoreUnknownFields);
+        }
+      } catch (IllegalArgumentException ex) {
+        throw ex;
```
gnanda, line 233: nit: Add a comment that this is for propagating errors from nested fields lower down the recursive stack

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -221,11 +221,22 @@ private static DynamicMessage convertJsonToProtoMessageImpl(
...
-            protoMsg, field, fieldSchema, json, jsonName, currentScope, ignoreUnknownFields);
+      try {
+        if (!field.isRepeated()) {
+          fillField(
+              protoMsg, field, fieldSchema, json, jsonName, currentScope, ignoreUnknownFields);
+        } else {
+          fillRepeatedField(
+              protoMsg, field, fieldSchema, json, jsonName, currentScope, ignoreUnknownFields);
+        }
+      } catch (IllegalArgumentException ex) {
```
gnanda, line 232: Should we instead create our own exception type here?

Then we can say "catch (OurException) { throw } catch (Exception) {wrap with OurException}", and IIUC this would let us remove all the try-catches inside fillField since this would add the field name to all errors.

gnanda (approved):
Approving for now in case we need to get them unblocked, but ideally the other way should be easier to maintain

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -225,7 +225,7 @@ public static class AppendSerializtionError extends RuntimeException {
     private final String streamName;
 
     public AppendSerializtionError(String streamName, Map<Integer, String> rowIndexToErrorMessage) {
```
justin-zhang-exa, line 227: Hi, I just noticed that this class name is also spelled incorrectly.

yirutang: Err... too bad. Maybe we will live with it since the error is already out.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
yirutang: Good idea, done. The only problem is maybe the info will be duplicated, but it is a cleaner solution.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
yirutang: Done.

PR #1796, opened 2022-09-23
chore: add test to check newBuilder for JsonStreamWriter

Neenu1995:
End to end tests added to test the [newBuilder() method](https://github.com/googleapis/java-bigquerystorage/pull/1794) added to JsonStreamWriter which enables users to create JsonStreamWriter without explicitly providing a Table Schema.

This PR has the following tests:
- Test to check that the JsonStreamWriter generated with the new methods works as expected.
- Test to check that the JsonStreamWriter generated with the newBuilder method is working exactly same as the JsonStreamWriter generated with older NewBuilder method by explicitly providingTableSchema.

### DIFF
--- a/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
+++ b/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
@@ -314,6 +314,161 @@ public class ITBigQueryWriteManualClientTest {
   }
 
   @Test
+  public void testJsonStreamWriterWithDefaultSchema()
+      throws IOException, InterruptedException, ExecutionException,
+          Descriptors.DescriptorValidationException {
+    String tableName = "JsonTableDefaultSchema";
+    TableFieldSchema TEST_STRING =
+        TableFieldSchema.newBuilder()
+            .setType(TableFieldSchema.Type.STRING)
+            .setMode(TableFieldSchema.Mode.NULLABLE)
+            .setName("test_str")
+            .build();
+    TableFieldSchema TEST_NUMERIC =
+        TableFieldSchema.newBuilder()
+            .setType(TableFieldSchema.Type.NUMERIC)
+            .setMode(TableFieldSchema.Mode.REPEATED)
+            .setName("test_numerics")
+            .build();
+    TableFieldSchema TEST_DATE =
+        TableFieldSchema.newBuilder()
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -468,6 +433,172 @@ public void testJsonStreamWriterWithDefaultSchema()
...
+      LOG.info("Sending one more message");
+      ApiFuture<AppendRowsResponse> response4 = jsonStreamWriter1.append(jsonArr4, -1);
+
+      Assert.assertFalse(response1.get().getAppendResult().hasOffset());
+      Assert.assertFalse(response2.get().getAppendResult().hasOffset());
+      Assert.assertFalse(response3.get().getAppendResult().hasOffset());
+      Assert.assertFalse(response4.get().getAppendResult().hasOffset());
+
+      LOG.info("Both JsonStreamWriters wrote to the same Table successfully!");
+    } catch (DescriptorValidationException e) {
```
yirutang, line 589: Will the test be passing with these exceptions? We should fail the test.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -468,6 +433,172 @@ public void testJsonStreamWriterWithDefaultSchema()
     }
   }
 
+  @Test
+  public void testMultipleJsonStreamWriterWithDefaultSchema()
```
yirutang, line 437: I think we only need one test with schema and one test without schema.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
Neenu1995: Added test for failure.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
Neenu1995: Removed multiple test.

yirutang (approved):

PR #1797, opened 2022-09-23
feat: client unknown fields drives writer refreshment

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -152,6 +152,9 @@ public class JsonStreamWriter implements AutoCloseable {
         LOG.warning("First attempt failed, waiting for 30 seconds to retry");
         Thread.sleep(30100);
         writeStream = client.getWriteStream(writeStreamRequest);
+        // TODO(yiru): We should let TableSchema return a timestamp so that we can simply
+        //     compare the timestamp to see if the table schema is the same. If it is the
+        //     same, we don't need to go refresh the writer again.
         refreshWriter(writeStream.getTableSchema());
         return JsonToProtoMessage.convertJsonToProtoMessage(
             this.descriptor, this.tableSchema, json, ignoreUnknownFields);
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -150,7 +150,7 @@ public class JsonStreamWriter implements AutoCloseable {
             this.descriptor, this.tableSchema, json, ignoreUnknownFields);
       } catch (Exceptions.JsonDataHasUnknownFieldException exex) {
         LOG.warning("First attempt failed, waiting for 30 seconds to retry");
-        Thread.sleep(30000);
+        Thread.sleep(30100);
         writeStream = client.getWriteStream(writeStreamRequest);
         refreshWriter(writeStream.getTableSchema());
         return JsonToProtoMessage.convertJsonToProtoMessage(
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -108,6 +108,43 @@ public class JsonStreamWriter implements AutoCloseable {
     return append(jsonArr, -1);
   }
 
+  private void refreshWriter(TableSchema updatedSchema)
+      throws DescriptorValidationException, IOException {
+    Preconditions.checkNotNull(updatedSchema, "updatedSchema is null.");
+    LOG.info("Refresh internal writer due to schema update");
+    // Close the StreamWriter
+    this.streamWriter.close();
+    // Update JsonStreamWriter's TableSchema and Descriptor
+    this.tableSchema = updatedSchema;
+    this.descriptor =
+        BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
+    this.protoSchema = ProtoSchemaConverter.convert(this.descriptor);
+    this.totalMessageSize = protoSchema.getSerializedSize();
+    // Create a new underlying StreamWriter with the updated TableSchema and Descriptor
+    this.streamWriter = streamWriterBuilder.setWriterSchema(this.protoSchema).build();
+  }
+
+  private Message buildMessage(JSONObject json)
+      throws InterruptedException, DescriptorValidationException, IOException {
+    try {
+      return JsonToProtoMessage.convertJsonToProtoMessage(
+          this.descriptor, this.tableSchema, json, ignoreUnknownFields);
+    } catch (Exceptions.JsonDataHasUnknownFieldException ex) {
+      // Backend cache for GetWriteStream schema staleness can be 30 seconds, wait a bit before
+      // trying to get the table schema to increase the chance of succeed. This is to avoid
+      // client's invalid data caused storm of GetWriteStream.
+      Thread.sleep(3500);
+      GetWriteStreamRequest writeStreamRequest =
+          GetWriteStreamRequest.newBuilder()
+              .setName(this.streamName)
+              .setView(WriteStreamView.FULL)
+              .build();
+      WriteStream writeStream = client.getWriteStream(writeStreamRequest);
+      refreshWriter(writeStream.getTableSchema());
+      return JsonToProtoMessage.convertJsonToProtoMessage(
+          this.descriptor, this.tableSchema, json, ignoreUnknownFields);
+    }
+  }
   /**
    * Writes a JSONArray that contains JSONObjects to the BigQuery table by first converting the JSON
    * data to protobuf messages, then using StreamWriter's append() to write the data at the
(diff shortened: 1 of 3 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -105,6 +105,43 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr)
...
+  private Message buildMessage(JSONObject json)
+      throws InterruptedException, DescriptorValidationException, IOException {
+    try {
+      return JsonToProtoMessage.convertJsonToProtoMessage(
+          this.descriptor, this.tableSchema, json, ignoreUnknownFields);
+    } catch (Exceptions.JsonDataHasUnknownFieldException ex) {
+      // Backend cache for GetWriteStream schema staleness can be 30 seconds, wait a bit before
+      // trying to get the table schema to increase the chance of succeed. This is to avoid
+      // client's invalid data caused storm of GetWriteStream.
+      Thread.sleep(3500);
```
dani29, line 133: I don't have enough understanding of the underlying staleness issue, but my hunch is that explicit Thread.sleep() within the library is going to mask plenty of weird errors. 

It's a bigger change, but consider allowing the user to supply a RetryPolicy as part of the builder, and then execute this convertJsonToProtoMessage() based on the user supplied policy. It gives the users more flexibility to iterate over different behaviors, i.e try more than just once, supply hard timeout, etc.

If this is out of scope in your opinion, then at least I would expect to see a LOG.warning before entering 3.5seconds of sleep.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -164,6 +190,8 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
           } else {
             rowIndexToErrorMessage.put(i, exception.getMessage());
           }
+        } catch (InterruptedException ex) {
+          throw new RuntimeException(ex);
```
dani29, line 194: AFAIU this is changing the behavior for downstream consumers who may handle exceptions differently, like Datastream, as now the exception is wrapped. please consult with @idobar on whether this affecting us.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -105,6 +105,47 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr)
...
+    } catch (Exceptions.JsonDataHasUnknownFieldException ex) {
+      // Backend cache for GetWriteStream schema staleness can be 30 seconds, wait a bit before
+      // trying to get the table schema to increase the chance of succeed. This is to avoid
+      // client's invalid data caused storm of GetWriteStream.
+      LOG.warning(
+          "Saw Json unknown field "
+              + ex.getFieldName()
+              + ", try to refresh the writer with updated schema");
+      Thread.sleep(35000);
+      GetWriteStreamRequest writeStreamRequest =
```
GaoleMeng, line 138: This seems problematic to me, what if after the sleep the updated schema is still not presented? And also any unknown field would cause customer's pipeline to delay 35 seconds?

yirutang: Maybe a new option for them to explicitly opt in? Usually, if they want to tolerant unknown fields, they could use ignoreUnknownFields and they will never see this error.

yirutang: Discussed offline, added an immediate retry to reduce the possible 30 second delay.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: I added one immediate retry and then a second try after 30 seconds, so that it can cover most of the case, while still delay bogus user ingestion.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -108,6 +109,58 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr)
     return append(jsonArr, -1);
   }
 
+  private void refreshWriter(TableSchema updatedSchema)
+      throws DescriptorValidationException, IOException {
+    Preconditions.checkNotNull(updatedSchema, "updatedSchema is null.");
+    LOG.info("Refresh internal writer due to schema update");
```
GaoleMeng, line 115: can you also log the stream name of the json writer?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -108,6 +109,58 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr)
...
+              .setView(WriteStreamView.FULL)
+              .build();
+      WriteStream writeStream = client.getWriteStream(writeStreamRequest);
+      refreshWriter(writeStream.getTableSchema());
+      try {
+        return JsonToProtoMessage.convertJsonToProtoMessage(
+            this.descriptor, this.tableSchema, json, ignoreUnknownFields);
+      } catch (Exceptions.JsonDataHasUnknownFieldException exex) {
+        LOG.warning("First attempt failed, waiting for 30 seconds to retry");
+        Thread.sleep(30100);
```
GaoleMeng, line 153: can be a global static constant

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -108,6 +109,58 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr)
...
+        BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
+    this.protoSchema = ProtoSchemaConverter.convert(this.descriptor);
+    this.totalMessageSize = protoSchema.getSerializedSize();
+    // Create a new underlying StreamWriter with the updated TableSchema and Descriptor
+    this.streamWriter = streamWriterBuilder.setWriterSchema(this.protoSchema).build();
+  }
+
+  private Message buildMessage(JSONObject json)
+      throws InterruptedException, DescriptorValidationException, IOException {
+    try {
```
GaoleMeng, line 130: Can you change this function structure to

```suggestion
// Try four times, each time 10 seconds wait if not with the first retry
for (int i = 0; i < 5; i++) {
  try {
    JsonToProtoMessage.convertJsonToProtoMessage(
          this.descriptor, this.tableSchema, json, ignoreUnknownFields);
    break;
  } catch (Exceptions.JsonDataHasUnknownFieldException ex) {
    if (i == 4) {
      // No need to try last iteration
      break;
    }
    if (i > 0) {
      // Sleep 10 seconds if not with the first retry.
      sleep(10000);
    }
    GetWriteStreamRequest writeStreamRequest =
                  GetWriteStreamRequest.newBuilder()
              .setName(this.streamName)
              .setView(WriteStreamView.FULL)
              .build();
     WriteStream writeStream = client.getWriteStream(writeStreamRequest);
     refreshWriter(writeStream.getTableSchema());
  }
}
```

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -127,16 +180,9 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
...
-        this.streamWriter.close();
-        // Update JsonStreamWriter's TableSchema and Descriptor
-        this.tableSchema = updatedSchema;
-        this.descriptor =
-            BQTableSchemaToProtoDescriptor.convertBQTableSchemaToProtoDescriptor(updatedSchema);
-        this.protoSchema = ProtoSchemaConverter.convert(this.descriptor);
-        this.totalMessageSize = protoSchema.getSerializedSize();
-        // Create a new underlying StreamWriter with the updated TableSchema and Descriptor
-        this.streamWriter = streamWriterBuilder.setWriterSchema(this.protoSchema).build();
+        if (updatedSchema != null) {
```
GaoleMeng, line 183: this is not necessary, we already checked this.streamWriter.getUpdatedSchema() != null

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: Each GetWriteStream call will end up as a RPC to our backend and I don't want to add too much pressure to our servers in the case of a user side bogus field. Imagine, the user has 1K-100K streams with a bogus field constantly hitting us. Most of the resolvable case should be able to be addressed by the immediate GetWriteStream call, if that GetWriteStream call doesn't work out, I only want to retry once after a period that I know for sure the issue should be gone and error out. I want to retry as little as possible, in the error case, I don't mind the user get stuck.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: Explained above.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
GaoleMeng: Can we change the iteration in my code from 5 to 3 in this case?

```suggestion
for (int i = 0; i < 3; i++) {
  try {
    JsonToProtoMessage.convertJsonToProtoMessage(
          this.descriptor, this.tableSchema, json, ignoreUnknownFields);
    break;
  } catch (Exceptions.JsonDataHasUnknownFieldException ex) {
    if (i == 3) {
      // No need to try last iteration
      break;
    }
    if (i == 1) {
      // Sleep 10 seconds if not with the first retry.
      sleep(30000);
    }
    GetWriteStreamRequest writeStreamRequest =
                  GetWriteStreamRequest.newBuilder()
              .setName(this.streamName)
              .setView(WriteStreamView.FULL)
              .build();
     WriteStream writeStream = client.getWriteStream(writeStreamRequest);
     refreshWriter(writeStream.getTableSchema());
  }
}

```

yirutang: No, I don't think it worth 3 times retry. I only want it to retry once.

GaoleMeng (approved):

PR #1798, opened 2022-09-23
feat: some fixes for multiplexing client 

GaoleMeng:
1. change setMinConnectionsPerPool -> setMinConnectionsPerRegion
2. disable logic for updated schema in multiplexing client 
3. close the passed in client if trying to share the same pool 
4. modify the header of the client before creating connection worker using the first request's write stream 


yirutang (approved):
Nit: We are calling GetWriteStream in our new builder and we can set location from that call there, but let's leave it to another CL.

PR #1802, opened 2022-09-26
feat: populate location info if we already called GetWriteStream

yirutang:
This PR is about if JsonStreamWriter already calls GetWriteStream to figure out the table's schema (should be what most users will use in the future), also populate location info from the result so that later on, we don't need to look up the location info again. Also added location accessor on StreamWriter and JsonStreamWriter.


GaoleMeng (commented):
Is the title of the PR correct? It's not introducing cache?

GaoleMeng (approved):

PR #1804, opened 2022-09-28
feat: add cache for location in stream writer, and trigger that when location is not presented

GaoleMeng:


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -43,6 +46,12 @@
 public class StreamWriter implements AutoCloseable {
   private static final Logger log = Logger.getLogger(StreamWriter.class.getName());
 
+  private static String datasetsMatching = "projects/[^/]+/datasets/[^/]+/";
+  private static Pattern streamPattern = Pattern.compile(datasetsMatching);
+
+  // Cache of location info for a given dataset.
+  private static Map<String, String> projectAndDatasetToLocation = new ConcurrentHashMap<>();
```
yirutang, line 53: Should we have a size limit of this cache? And also probably an expiration time. I am not sure how often is this but the user can delete a dataset and recreate one in different region?

yirutang: Let's just add a size limit for now and add expiration later.

GaoleMeng: I didn't add a size limit because it's plain string, seems not possible to exceed limit. We can cache millions of entries without problem I think and it's kinda hard to imagine we need to cache that many..

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
yirutang: Sure fine, let's first go with this. I am a bit worry about dataset recreate case but that should be a corner case. Let's get going.

PR #1813, opened 2022-10-03
fix: return row-level error information via an AppendSerializtionError

agrawal-siddharth:
exception.

feat: return list of row errors from append

### DIFF
--- a/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
+++ b/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
@@ -28,6 +28,7 @@ import com.google.cloud.bigquery.*;
 import com.google.cloud.bigquery.Schema;
 import com.google.cloud.bigquery.storage.test.Test.*;
 import com.google.cloud.bigquery.storage.v1.*;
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -313,6 +314,73 @@ public void testJsonStreamWriterCommittedStream()
...
+                    WriteStream.newBuilder().setType(WriteStream.Type.COMMITTED).build())
+                .build());
+    StreamWriter streamWriter =
+        StreamWriter.newBuilder(writeStream.getName())
+            .setWriterSchema(ProtoSchemaConverter.convert(FooType.getDescriptor()))
+            .build();
+    LOG.info("Sending three messages");
+    ApiFuture<AppendRowsResponse> futureResponse =
+        streamWriter.append(
+            CreateProtoRows(new String[] {"aaabbbcccddd", "bbb", "cccdddeeefffggg"}), -1);
```
yirutang, line 348: Shouldn't block this checkin, could you try to hit the timestamp error in the bug to verify that path is fixed?

yirutang (approved):

yirutang (commented):
Some of the presubmit checks are failing, such as the title of the PR needs to be:
fix: xxxx

PR #1822, opened 2022-10-04
fix: breaking change to continue supporting users who may be

agrawal-siddharth:
consuming errors as StatusRuntimeException

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -217,15 +217,20 @@ public final class Exceptions {
 
   /**
    * This exception is thrown from {@link JsonStreamWriter#append()} when the client side Json to
-   * Proto serializtion fails. The exception contains a Map of indexes of faulty lines and the
-   * corresponding error message.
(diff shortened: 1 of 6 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -217,15 +217,20 @@ public static StorageException toStorageException(Throwable exception) {
...
    */
-  public static class AppendSerializtionError extends RuntimeException {
+  public static class AppendSerializtionError extends StatusRuntimeException {
     private final Map<Integer, String> rowIndexToErrorMessage;
     private final String streamName;
 
-    public AppendSerializtionError(String streamName, Map<Integer, String> rowIndexToErrorMessage) {
-      super(String.format("Append serialization failed for writer: %s", streamName));
+    public AppendSerializtionError(
+        int codeValue,
```
yirutang, line 229: why not directly accept a Status? So for the response callback case, you don't need to disassemble it and assemble it.

yirutang: nvm

yirutang (approved):

PR #1841, opened 2022-10-21
fix: add multiplexing e2e test with client shutdown

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
+++ b/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
@@ -62,15 +62,24 @@ public class ITBigQueryWriteManualClientTest {
   private static final String DATASET = RemoteBigQueryHelper.generateDatasetName();
   private static final String DATASET_EU = RemoteBigQueryHelper.generateDatasetName();
   private static final String TABLE = "testtable";
-  private static final String TABLE2 = "complicatedtable";
+  private static final String TABLENESTED = "complicatedtable";
   private static final String DESCRIPTION = "BigQuery Write Java manual client test dataset";
 
   private static BigQueryWriteClient client;
   private static TableInfo tableInfo;
   private static TableInfo tableInfo2;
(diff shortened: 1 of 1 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -1227,33 +1258,37 @@ public void testOffsetOutOfRangeError() throws IOException, InterruptedException
     }
   }
 
+  void writeSomeRows(StreamWriter sw, List<Future<AppendRowsResponse>> futureList) {
```
alvarowolfx, line 1261: nit: change this function to be :

```
void writeRows(StreamWriter sw, numRows int, List<Future<AppendRowsResponse>> futureList) {
  for (int i = 0; i < numRows; i++) {
    sw.append(CreateProtoRows(new String[] {"aaa", "bbb", "ccc"}));
  }
}
```

So it's a bit more reusable and add meaning to it instead of using the word "some". First time I read I asked myself: "How many rows this is inserting ?"

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -105,17 +124,31 @@ public static void beforeClass() throws IOException {
                             .setMode(Field.Mode.NULLABLE)
                             .build())))
             .build();
+    bigquery.create(tableInfo);
```
alvarowolfx, line 127: Seems like all the 20 tables can be created with the `createTableHelper(i)` call, instead of creating the first one separately and later creating 19 more.

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
yirutang: d.

alvarowolfx (approved):

PR #1863, opened 2022-10-31
feat: provide sample code for row-level error handling

agrawal-siddharth:


samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -191,7 +197,37 @@ public void onFailure(Throwable throwable) {
...
+              }
+            }
+
+            // Mark the existing attempt as done since we got a response for it
+            done();
+
+            // Retry the remaining valid rows.
+            if (dataNew.length() > 0) {
+              try {
+                this.parent.append(new AppendContext(dataNew, 0));
```
yirutang, line 224: Is there a way to start another thread and do this append? Currently this append can block waiting for inflight limit...

samples/snippets/src/test/java/com/example/bigquerystorage/WriteToDefaultStreamIT.java
```diff
@@ -73,7 +73,11 @@ public void setUp() {
     // Create a new dataset and table for each test.
     datasetName = "WRITE_STREAM_TEST" + UUID.randomUUID().toString().substring(0, 8);
     tableName = "DEFAULT_STREAM_TEST" + UUID.randomUUID().toString().substring(0, 8);
-    Schema schema = Schema.of(Field.of("test_string", StandardSQLTypeName.STRING));
+    Schema schema = Schema.of(
```
yirutang, line 76: Where is the actual test?

agrawal-siddharth: https://github.com/googleapis/java-bigquerystorage/blob/main/samples/snippets/src/test/java/com/example/bigquerystorage/WriteToDefaultStreamIT.java

has the function testWriteToDefaultStream(). This calls WriteToDefaultStream.writeToDefaultStream() which does the test.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
agrawal-siddharth: Done. Both retries and row-errors are now handled by calls to append() made from a separate thread.

yirutang (approved):

PR #1867, opened 2022-11-05
feat: add update schema support for multiplexing

GaoleMeng:
To make this happen, we will store a mapping from stream name to updated schema mapping inside connection worker pool. Whenever the json writer accept one append, we will check the cache to see whether there is one updated schema and compared with the current one. Then recreate the stream writer if there is different schema


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -421,4 +437,23 @@ FlowController.LimitExceededBehavior limitExceededBehavior() {
...
+    }
+
+    @Override
+    public void onFailure(Throwable t) {
+      // No Op during failure.
+    }
+
+    @Override
+    public void onSuccess(AppendRowsResponse response) {
+      streamNameToUpdatedSchema.put(streamName, response.getUpdatedSchema());
```
yirutang, line 456: After you refreshWriter, you need to mark the entry here as null. I think it is better we keep a schema on the Writer level.

GaoleMeng: The problem is there could be multiple stream writers that use the same stream name that all need refreshWriter to be triggered whenever there is a updated schema

So we can't directly nullify the updated schema for a given stream name, otherwise some streamwriter might not be able to get the updated schema correctly.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -187,9 +186,9 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
     // Handle schema updates in a Thread-safe way by locking down the operation
     synchronized (this) {
       // Update schema only work when connection pool is not enabled.
-      if (this.streamWriter.getConnectionOperationType() == Kind.CONNECTION_WORKER
-          && this.streamWriter.getUpdatedSchema() != null) {
-        refreshWriter(this.streamWriter.getUpdatedSchema());
+      TableSchema updatedSchema = this.streamWriter.getUpdatedSchema();
+      if (updatedSchema != null && !this.tableSchema.equals(updatedSchema)) {
```
yirutang, line 190: Will this.tableSchema be updated?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -187,9 +186,9 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
...
+        refreshWriter(updatedSchema);
```
yirutang, line 191: Per discussion, we only need to update the tableSchema and be sure to send in the new schema, without needing to reconnect?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: nmv, refreshWriter should just work here.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yirutang: nvm.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -398,7 +397,7 @@ public static StreamWriter.Builder newBuilder(String streamName) {
 
   /** Thread-safe getter of updated TableSchema */
   public synchronized TableSchema getUpdatedSchema() {
-    return singleConnectionOrConnectionPool.getUpdatedSchema();
+    return singleConnectionOrConnectionPool.getUpdatedSchema(this);
```
yirutang, line 400: There is a little meaning change in this field now. Previously, it will only return an Updated Schema when a schema update happens during the life time of this StreamWriter. Now it will always return the "current schema" of our knowledge. May worth explain this a bit since Dataflow is going to use this field.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -65,6 +68,11 @@ public class ConnectionWorkerPool {
   private final Set<ConnectionWorker> connectionWorkerPool =
       Collections.synchronizedSet(new HashSet<>());
 
+  /*
+   * Contains the mapping from stream name to updated schema.
+   */
+  private Map<String, TableSchema> streamNameToUpdatedSchema = new ConcurrentHashMap<>();
```
yirutang, line 74: Is there a way to only cache to the level of table, the size of this map could be huge, if it is per stream one table schema.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
GaoleMeng: Added one line comment

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
GaoleMeng: Changed to cache table name

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
GaoleMeng: As discussed offline, changed to use timestamp pattern 

If the timestamp used in the current stream writer is older than the updated schema version,change to use updated schema

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -720,7 +722,7 @@ private AppendRequestAndResponse pollInflightRequestQueue() {
   }
 
   /** Thread-safe getter of updated TableSchema */
-  public synchronized TableSchema getUpdatedSchema() {
+  public synchronized TableSchemaAndTimestamp getUpdatedSchema() {
```
yirutang, line 725: This is breaking change. Lets just add a new method instead of change this old method.

GaoleMeng: This method should not been used as public,
let's fallback to package private

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -147,11 +152,11 @@ long getInflightWaitSeconds(StreamWriter streamWriter) {
       return connectionWorker().getInflightWaitSeconds();
     }
 
-    TableSchema getUpdatedSchema() {
+    TableSchemaAndTimestamp getUpdatedSchema(StreamWriter streamWriter) {
```
yirutang, line 155: Sorry, I mean this is actually a breaking change? Dataflow will use this method.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
yirutang: Yeah, I agree that for ConnectionWorker this method shouldn't be public at all. But it seems the method on StreamWriter also changed?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
GaoleMeng: Discussed offline, let's used timestamp on streamwriter when returning schema

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -186,10 +185,10 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
       throws IOException, DescriptorValidationException {
     // Handle schema updates in a Thread-safe way by locking down the operation
     synchronized (this) {
-      // Update schema only work when connection pool is not enabled.
-      if (this.streamWriter.getConnectionOperationType() == Kind.CONNECTION_WORKER
-          && this.streamWriter.getUpdatedSchema() != null) {
-        refreshWriter(this.streamWriter.getUpdatedSchema());
+      TableSchema updatedSchemaAndTime = this.streamWriter.getUpdatedSchema();
+      // Create a new stream writer internally if a new updated schema is reported from backend.
+      if (updatedSchemaAndTime != null && !this.tableSchema.equals(updatedSchemaAndTime)) {
```
yirutang, line 190: We can directly use streamWriter.getUpdatedSchema() != null?

GaoleMeng: Done

yirutang (approved):

PR #1884, opened 2022-11-15
fix: Fix window build bug caused by Instant resolution.

GaoleMeng:
The bug happened because Windows Instant resolution is about 10ms level:
https://stackoverflow.com/a/20689231
We decided to change test directly, as in reality this bug can seldom happened due to the updated schema is required to happen after a server/client communication, which is at 100ms level 


yirutang (approved):

Neenu1995 (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -186,7 +186,9 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
     // Handle schema updates in a Thread-safe way by locking down the operation
     synchronized (this) {
       // Create a new stream writer internally if a new updated schema is reported from backend.
-      if (this.streamWriter.getUpdatedSchema() != null) {
+      TableSchema updatedSchema = this.streamWriter.getUpdatedSchema();
+      // Only refresh when the reference of the schema is different.
+      if (updatedSchema != null && this.tableSchema != updatedSchema) {
```
yirutang, line 191: Seems this is just for the extreme case of if a schema update event happens within 10ms of Stream creation. I think we should be OK without this comparison.

GaoleMeng: Discussed offline, changed to change test only

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -185,6 +185,7 @@ public ApiFuture<AppendRowsResponse> append(JSONArray jsonArr, long offset)
       throws IOException, DescriptorValidationException {
     // Handle schema updates in a Thread-safe way by locking down the operation
     synchronized (this) {
+      TableSchema updatedSchemaAndTime = this.streamWriter.getUpdatedSchema();
```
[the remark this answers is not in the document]
GaoleMeng: Done

yirutang (approved):

PR #1930, opened 2023-01-13
fix: add a timeout on retry for retryable errors

yirutang:

Fixes 1931


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -686,15 +697,23 @@ private void doneCallback(Throwable finalStatus) {
...
       if (connectionFinalStatus == null) {
+        if (connectionRetryStartTime == 0) {
+          connectionRetryStartTime = System.currentTimeMillis();
+        }
         // If the error can be retried, don't set it here, let it try to retry later on.
-        if (isRetriableError(finalStatus) && !userClosed) {
+        if (isRetriableError(finalStatus)
+            && !userClosed
+            && System.currentTimeMillis() - connectionRetryStartTime
+                <= maxRetryDuration.toMillis()) {
```
shollyman, line 707: What if the user doesn't want a max retry duration enforced?  Should there be special handling for the zero value?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/FakeBigQueryWrite.java
```diff
@@ -91,6 +91,10 @@ public void setTimesToClose(long numberTimesToClose) {
     serviceImpl.setTimesToClose(numberTimesToClose);
   }
 
+  public void setCloseForeverAfter(long closeForeverAfter) {
```
shollyman, line 94: I get this is a fake, but I'm not clear from the name what it's purpose is.  A better name might make sense here or a comment clarifying usage.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
yirutang: Sure.

shollyman (approved):

PR #1933, opened 2023-01-17
fix: we should isolate the client used in StreamWriter and the client used in ConnectionWorker

yirutang:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
@@ -207,8 +207,8 @@ public class StreamWriter implements AutoCloseable {
 
       boolean ownsBigQueryWriteClient = builder.client == null;
       // We need a client to perform some getWriteStream calls.
-      BigQueryWriteClient client = builder.client != null ?
-          builder.client : new BigQueryWriteClient(clientSettings);
+      BigQueryWriteClient client =
+          builder.client != null ? builder.client : new BigQueryWriteClient(clientSettings);
       String location = builder.location;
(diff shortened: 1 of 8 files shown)


GaoleMeng (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -206,7 +205,9 @@ private StreamWriter(Builder builder) throws IOException {
             "Trying to enable connection pool in non-default stream.");
       }
 
-      BigQueryWriteClient client = getBigQueryWriteClient(builder);
+      // We need a client to perform some getWriteStream calls.
+      BigQueryWriteClient client =
```
GaoleMeng, line 209: why do we need to create another client here if the passed in client is null? 

we won't use the created client

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -290,19 +288,17 @@ static boolean isDefaultStream(String streamName) {
...
-      BigQueryWriteSettings stubSettings =
-          BigQueryWriteSettings.newBuilder()
-              .setCredentialsProvider(builder.credentialsProvider)
-              .setTransportChannelProvider(builder.channelProvider)
-              .setBackgroundExecutorProvider(builder.executorProvider)
-              .setEndpoint(builder.endpoint)
-              .build();
-      testOnlyClientCreatedTimes++;
-      return BigQueryWriteClient.create(stubSettings);
+  private BigQueryWriteSettings getBigQueryWriteSettings(
```
GaoleMeng, line 291: Only pass in builder is enough?

PR #1938, opened 2023-01-18
fix: fix deadlock issue in ConnectionWorkerPool

GaoleMeng:


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPoolTest.java
```diff
@@ -315,6 +321,68 @@ public void testMultiStreamAppend_appendWhileClosing() throws Exception {
     assertThat(connectionWorkerPool.getTotalConnectionCount()).isEqualTo(0);
   }
 
+  @Test
+  public void testBlocking() throws Exception {
```
yirutang, line 325: Is this test hitting the issue and how? Clean up the test?

yirutang (approved):

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPoolTest.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -623,6 +662,11 @@ private void requestCallback(AppendRowsResponse response) {
       }
       if (!this.inflightRequestQueue.isEmpty()) {
         requestWrapper = pollInflightRequestQueue();
+
+        Instant insertTime = requestWrapper.insertTime;
+        Instant currentTime = Instant.now();
+        long latencyToAdd = currentTime.toEpochMilli() - insertTime.toEpochMilli();
+        updateLatencyDistribution(latencyToAdd);
```
yirutang, line 669: Let's add the error message for now from client lib at L686. So in case there is still a blockage on customer's case, we will have more info to investigate.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -234,7 +251,26 @@ public void run() {
...
+                while (true) {
+                  reportSingleCurrentStatus();
+                }
+              }
+            });
     this.appendThread.start();
+    this.reportingThread.start();
+  }
+
+  private void reportSingleCurrentStatus() {
```
yirutang, line 269: Is this too much logging? Is it going to be in the final code?

GaoleMeng: Not going to be in the final code, it's reported every 5 seconds

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Do you mean the error message in response? I think that's already printed by Pablo's current code... I was planning to revert the change in this file once Pablo copy over this part

yirutang: Yes, we want it to be in the release we hand over to customers, so if anything is going on, we have some logs.

GaoleMeng: let me clean this file up, Pablo has already copied the stats log, let's add error log directly

GaoleMeng: But won't that kind of log overwhelmed the log console?

PR #1943, opened 2023-01-19
Some refactoring ideas for Storage client library

reuvenlax:
Remove streamWriterToConnection and connectionToWriteStream maps, and instead store this data in the StreamWriter and ConnectionWorker objects themselves. This means that we no longer have to do a map lookup on every call to append().

Instead of using timestamps to determine which StreamWriter objects to send updated schema to, use a registration method. This way only StreamWriters that were created prior to a schema-update callback will get the updated schema (Note: this uses a static map, but could instead be done by updating the StreamWriter directly). I think this preserves the intended semantics from before, but needs a good look. Note: the timestamp approach isn't guaranteed to work, since it's possible for the time to stay the same between StreamWriter creation and the callback (Java does not guarantee that System.nanoTime() actually updates every nanosecond - it provides no guarantee on update frequency).

Some other things worth experimenting with in the future:
   - See if we can remove the global lock in ConnectionWorkerPool and instead use local locks in StreamWriter and ConnectionWorker. This might reduce lock contention, but will cause us to always grab two locks instead of 1 which might use more CPU. Unclear which approach is better.
   -  Right now ConnectionWorkerPool always checks isOverwhelmed on every call to append and looks for a new stream in that case. If we're in a case where all streams are at their maximum, this might cause a lot of extra CPU usage at the time we can least afford it (when the worker is already overwhelmed!). We should consider throttling this - e.g. maybe we only move a StreamWriter to a new stream if it's been > 1 sec since the last time we checked.
   - As mentioned above, there are other ways of dealing with updatedSchema.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
@@ -20,7 +20,6 @@ import com.google.api.core.ApiFutures;
 import com.google.api.gax.batching.FlowController;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.bigquery.storage.v1.ConnectionWorker.Load;
-import com.google.cloud.bigquery.storage.v1.ConnectionWorker.TableSchemaAndTimestamp;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
@@ -29,6 +28,7 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -67,14 +67,6 @@ public class ConnectionWorkerPool {
    */
   private final FlowController.LimitExceededBehavior limitExceededBehavior;
 
-  /** Map from write stream to corresponding connection. */
-  private final Map<StreamWriter, ConnectionWorker> streamWriterToConnection =
-      new ConcurrentHashMap<>();
-
-  /** Map from a connection to a set of write stream that have sent requests onto it. */
-  private final Map<ConnectionWorker, Set<StreamWriter>> connectionToWriteStream =
-      new ConcurrentHashMap<>();
-
   /** Collection of all the created connections. */
   private final Set<ConnectionWorker> connectionWorkerPool =
       Collections.synchronizedSet(new HashSet<>());
@@ -82,7 +74,8 @@ public class ConnectionWorkerPool {
   /*
    * Contains the mapping from stream name to updated schema.
    */
-  private Map<String, TableSchemaAndTimestamp> tableNameToUpdatedSchema = new ConcurrentHashMap<>();
+  private Map<String, Set<StreamWriter>> streamNameToStreamWriter = new ConcurrentHashMap();
+  private Map<StreamWriter, TableSchema> streamWriterToUpdatedSchema = new ConcurrentHashMap<>();
 
   /** Enable test related logic. */
   private static boolean enableTesting = false;
@@ -225,6 +218,12 @@ public class ConnectionWorkerPool {
     ConnectionWorkerPool.settings = settings;
   }
 
+  public void registerStreamWriter(StreamWriter streamWriter) {
+    streamNameToStreamWriter
+        .computeIfAbsent(streamWriter.getStreamName(), k -> new HashSet<>())
+        .add(streamWriter);
+  }
+
   /** Distributes the writing of a message to an underlying connection. */
   public ApiFuture<AppendRowsResponse> append(StreamWriter streamWriter, ProtoRows rows) {
     return append(streamWriter, rows, -1);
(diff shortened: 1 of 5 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -234,47 +233,52 @@ public ApiFuture<AppendRowsResponse> append(StreamWriter streamWriter, ProtoRows
...
+        } catch (IOException e) {
+          throw new IllegalStateException(e);
+        }
+        currentConnection = createdOrExistingConnection;
+        streamWriter.setCurrentConnectionPoolConnection(currentConnection);
+        // Update connection to write stream relationship.
+        // TODO: What if we simply kept an atomic refcount in ConnectionWorker? We could also
+        // manage the refcount in the callback below to precisely track which connections are being
+        // used.
+        currentConnection.getCurrentStreamWriters().add(streamWriter);
```
yirutang, line 258: Is this lock good enough to protect currentConnection?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -234,47 +233,52 @@ public ApiFuture<AppendRowsResponse> append(StreamWriter streamWriter, ProtoRows
...
-                // Update connection to write stream relationship.
-                connectionToWriteStream.computeIfAbsent(
-                    createdOrExistingConnection, (ConnectionWorker k) -> new HashSet<>());
-                connectionToWriteStream.get(createdOrExistingConnection).add(streamWriter);
-                return createdOrExistingConnection;
-              } finally {
-                lock.unlock();
-              }
-            });
+    ConnectionWorker currentConnection;
```
yirutang, line 236: In general, I like this idea. We can reuse the connection more for the same StreamWriter.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -170,6 +184,11 @@ String getWriterId(String streamWriterId) {
       return connectionWorker().getWriterId();
     }
 
+    public void register(StreamWriter streamWriter) {
```
GaoleMeng, line 187: Can be removed?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -155,12 +169,12 @@ long getInflightWaitSeconds(StreamWriter streamWriter) {
       return connectionWorker().getInflightWaitSeconds();
     }
 
-    TableSchemaAndTimestamp getUpdatedSchema(StreamWriter streamWriter) {
-      if (getKind() == Kind.CONNECTION_WORKER_POOL) {
+    TableSchema getUpdatedSchema(StreamWriter streamWriter) {
+      if (getKind() == Kind.CONNECTION_WORKER) {
+        return connectionWorker().getUpdatedSchema();
+      } else {
         return connectionWorkerPool().getUpdatedSchema(streamWriter);
```
yirutang, line 176: Does this breaks the promise of StreamWriter only saw updates when there is a schema update? I think we should be fine to use nano time since it is monotonic on this machine?https://screenshot.googleplex.com/3Qeo9ouZEnehgMR

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -234,47 +233,52 @@ public ApiFuture<AppendRowsResponse> append(StreamWriter streamWriter, ProtoRows
...
+    // TODO: Do we need a global lock here? Or is it enough to just lock the StreamWriter?
+    lock.lock();
+    try {
+      currentConnection = streamWriter.getCurrentConnectionPoolConnection();
```
yirutang, line 240: Should we keep multiple (at least 2) connections in order to scale up, and avoid look into the global pool?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -234,47 +233,52 @@ public ApiFuture<AppendRowsResponse> append(StreamWriter streamWriter, ProtoRows
...
+        try {
+          createdOrExistingConnection =
+              createOrReuseConnectionWorker(streamWriter, currentConnection);
+        } catch (IOException e) {
+          throw new IllegalStateException(e);
+        }
+        currentConnection = createdOrExistingConnection;
+        streamWriter.setCurrentConnectionPoolConnection(currentConnection);
+        // Update connection to write stream relationship.
+        // TODO: What if we simply kept an atomic refcount in ConnectionWorker? We could also
```
GaoleMeng, line 255: refcount would be error prone as we streamwriter could be switching back and forth between connection workers meaning one worker could be recording a single stream writer multiple times if using refcount

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
reuvenlax: I don't understand the comment?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
reuvenlax: we're using it to protect currentConnection.getCurrentStreamWriters()

In theory we could put a lock inside of currentConnection, which would create more granular locking. However this would also cause a lot more lock/unlock activity (e.g. every call to append would have to lock at least two locks) so this change would need measurement to see if it better better.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
reuvenlax: I don't think nanoTime completely works. it is monotonic, but not strictly increasing - i.e. the current code is broken if the first update has the same nanoTime as the creation time, which is completely possible.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
reuvenlax: I think that would be fine. The refcount removal would happen in the done callback (below in ApiFutures.transform), so would know exactly which connection worker to decrement even if the stream writer has moved to a different stream.

PR #1954, opened 2023-01-23
fix: reduce visibility of the ConnectionPool and ConnectionWorker, so…

yirutang:
… we can reduce clirr error

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
@@ -227,8 +227,7 @@ public class ConnectionWorkerPool {
 
   /** Distributes the writing of a message to an underlying connection. */
   @VisibleForTesting
-  ApiFuture<AppendRowsResponse> append(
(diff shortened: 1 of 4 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerPool.java
```diff
@@ -192,7 +193,8 @@ public abstract static class Builder {
   /** Static setting for connection pool. */
   private static Settings settings = Settings.builder().build();
 
-  public ConnectionWorkerPool(
+  @VisibleForTesting
```
GaoleMeng, line 196: VisibleForTesting is for test only I think

yirutang: Yes. Other users will use this ctor? I thought they only need to use the settings on the pool?

yirutang: Done.

GaoleMeng (approved):

PR #1955, opened 2023-01-23
fix: fix one potential root cause of deadlock in connection worker

GaoleMeng:
The deadlock may happen when we try to insert messages inside message callback and the request queue happen to be full.
During the above mentioned pattern, the following happened in sequence:
1. RequestCallback triggered in connectionWorker.java 
2. appendResult.setException is triggered 
3. user's provided callback triggered, triggering append
4. inflight queue happen to be full, stuck in waiting in flight queue back in normal 
however inflight queue can't be progressed until step 1 finish, as there could be only one requestCallback triggered each time
The fix was to fix the example code by puting the append at another thread, it solved issue but not good enough. The true fix is to put step 2 to another thread directly so that step 1 is unblocked. This will fix any user on the original example code directly.

The example code can also be upgraded to remove the unnecessary thread pool


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -639,35 +642,44 @@ private void requestCallback(AppendRowsResponse response) {
...
-      }
-    } else {
-      requestWrapper.appendResult.set(response);
-    }
+
+    // We need a separte thread pool to unblock the next request callback.
+    // Otherwise user may call append inside request callback, which may be blocked on waiting
+    // on in flight quota, causing deadlock as requests can't be popped out of queue until
+    // the current request callback finishes.
+    threadPool.submit(
```
yirutang, line 650: I don't see how the code in this block can cause any blocks. It doesn't need a lock to execute and it is just some plain conversion code?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -63,6 +65,7 @@ public class ConnectionWorker implements AutoCloseable {
   private Condition hasMessageInWaitingQueue;
   private Condition inflightReduced;
   private static Duration maxRetryDuration = Duration.ofMinutes(5);
+  private static ExecutorService threadPool = Executors.newFixedThreadPool(50);
```
yirutang, line 68: Looks good, let's do 20.

yirutang (approved):

PR #1957, opened 2023-01-25
feat: add timeout to inflight queue waiting

GaoleMeng:
If we don't see the exception thrown by timeout on waiting, we then know it's not deadlock on waiting for in flight queue to have space 


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -387,6 +406,7 @@ public void close() {
     }
 
     try {
+      log.info("Begin shutting down user callback thread pool for stream " + streamName);
```
yirutang, line 409: Is this logging accurate? with multiplexing, the identifier shouldn't be streamName. It is not just for this change but in general, seems streamName is not good enough to indicate a connection.

Also seems we should change this log to fine level.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -397,6 +417,7 @@ public void close() {
               + " is interrupted with exception: "
               + e.toString());
     }
+    log.info("User close finishes for stream " + streamName);
```
yirutang, line 420: We can remove the line on 395?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -387,6 +406,7 @@ public void close() {
     }
 
     try {
+      log.info("Begin shutting down user callback thread pool for stream " + streamName);
       threadPool.shutdown();
       threadPool.awaitTermination(3, TimeUnit.MINUTES);
```
yirutang, line 411: Log when we actually timed out here?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -61,6 +61,9 @@
 public class ConnectionWorker implements AutoCloseable {
   private static final Logger log = Logger.getLogger(StreamWriter.class.getName());
 
+  // Maximum wait time on inflight quota before error out.
+  private static long INFLIGHT_QUOTA_MAX_WAIT_TIME_MILLI = 180000;
```
yirutang, line 65: Make the default to 5 minutes?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Yes, we need a new kind of identifier besides stream name 

let's change them together in the next PR

changed to fine

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Done

yirutang (approved):

PR #1964, opened 2023-01-27
feat: allow java client to handle schema change during same stream name 

GaoleMeng:
We will consider multiplexing to be entered during schema switch. 

Also fixed the clirrr error 


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -483,25 +483,31 @@ private void appendLoop() {
...
+        // a new schema for the same stream name.
+        // For the schema comparision we don't use message differencer to speed up the comparing
+        // process. `equals(...)` can bring us false positive, e.g. two repeated field can be
+        // considered the same but is not considered equals(). However as long as it's never provide
+        // false negative we will always correctly pass writer schema to backend.
+        if ((!originalRequest.getWriteStream().isEmpty()
             && !streamName.isEmpty()
-            && !originalRequest.getWriteStream().equals(streamName)) {
+            && !originalRequest.getWriteStream().equals(streamName))
+            || !originalRequest.getProtoRows().getWriterSchema().equals(writerSchema)) {
```
yirutang, line 501: could this be null?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
```diff
@@ -247,10 +247,10 @@ public void testAppendInSameStream_switchSchema() throws Exception {
         // We will get the request as the pattern of:
         // (writer_stream: t1, schema: schema1)
         // (writer_stream: _, schema: _)
-        // (writer_stream: _, schema: schema3)
```
yirutang, line 250: I am confused, so previously, just switch the schema works?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: No as we will always pass in write schema on the original request, but let's check that to make sure

yirutang (approved):

PR #1966, opened 2023-01-30
feat: add functions to set missing value map in the stream writers

risnayak:
We allow users to set a missing value map for the StreamWriter and JsonStreamWriter. The value of the missing value map will be re-used by the writer till changed otherwise.


yirutang (commented):
Looks good, could you add a e2e test to make sure it works?
https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java

risnayak:
> Looks good, could you add a e2e test to make sure it works? https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java

Yep I can definitely add E2E tests! The only thing issue is we'd first need https://github.com/googleapis/java-bigquerystorage/pull/1988 to be submitted, and then we can add E2E tests (since we won't be able to create tables with default values without this).

If it sounds ok, I can add the E2E test in a followup request.

yirutang (approved):

PR #1981, opened 2023-02-07
fix: refactor only, add StreamWriter to AppendRowsRequestResponse

yirutang:
1. The stream name of the writer can be used for separate queue processing when we start to add per table ordering.
2. It allows us to callback on streamwriter for its internal counter, so we can close the streamwriter with inflight requests done.


GaoleMeng (commented):
Why do we need streamwriter to the appendRowsRequestResponse?

yirutang:
> Why do we need streamwriter to the appendRowsRequestResponse?

As stated in description, there are two potential usage of this:
1. To callback on StreamWriter when we got a response, so that on SW close, we can wait for the count to go down to 0. I am going to send out a PR about this as next step.
2. When we add the appendRowsRequestResponse, we can use the StreamName to build up a separate queue for each different StreamName.

GaoleMeng (approved):

PR #1982, opened 2023-02-08
feat: Add sample to indicate how to recreate a stream writer.

yirutang:

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
+++ b/samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
@@ -158,18 +158,18 @@ public class WriteToDefaultStream {
     }
 
     public synchronized void append(AppendContext appendContext)
-        throws DescriptorValidationException, IOException {
+        throws DescriptorValidationException, IOException, InterruptedException {
       synchronized (this.lock) {
         // If earlier appends have failed, we need to reset before continuing.
         if (this.error != null) {
           throw this.error;
         }
       }
-      // Append asynchronously for increased throughput.
       if (streamWriter.isClosed() && !streamWriter.isUserClosed()) {
         streamWriter.close();
         initialize(streamWriter.getStreamName());
       }
+      // Append asynchronously for increased throughput.
       ApiFuture<AppendRowsResponse> future = streamWriter.append(appendContext.data);
       ApiFutures.addCallback(
           future, new AppendCompleteCallback(this, appendContext), MoreExecutors.directExecutor());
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -135,22 +135,29 @@ private static class DataWriter {
...
       // For more information about JsonStreamWriter, see:
       // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.html
       streamWriter =
-          JsonStreamWriter.newBuilder(parentTable.toString(), BigQueryWriteClient.create()).build();
+          JsonStreamWriter.newBuilder(streamOrTableName, BigQueryWriteClient.create())
+              .build();
     }
 
-    public void append(AppendContext appendContext)
+    public synchronized void append(AppendContext appendContext)
```
GaoleMeng, line 160: why this need to be changed to synchronized?

yirutang: Because we could be replacing the writer in the function. The other way is to make the writer be protected by a lock, but that would require writer.append be protected by a lock too.

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -135,22 +135,29 @@ private static class DataWriter {
...
+      initialize(parentTable.toString());
+    }
+
+    private void initialize(String streamOrTableName)
+        throws DescriptorValidationException, IOException, InterruptedException {
       // Use the JSON stream writer to send records in JSON format. Specify the table name to write
       // to the default stream.
       // For more information about JsonStreamWriter, see:
       // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.html
       streamWriter =
```
GaoleMeng, line 155: we don't need to call close() on the previous streamwriter?

yirutang: Yifat asked the same question, aka, stream worker will call close automatically on GC. 
https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L445

But I am also not against add one explicitly. WDYT?

GaoleMeng: GC will trigger much time later than this overwrite and GC won't care autoclosable (it won't trigger close() during GC)
https://stackoverflow.com/questions/8321767/closeable-in-garbage-collection

Autoclosable is explicitly used for try with resource which is not related to GC 
Let's close the previous one explicitly

yirutang: Good point!

samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
yirutang: I removed the synchronize since I found the append is already in a lock below.

PR #1984, opened 2023-02-08
fix: catch uncaught exception from append loop and add expoential retry to reconnection 

GaoleMeng:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [2.32.1](https://github.com/googleapis/java-bigquerystorage/compare/v2.32.0...v2.32.1) (2023-02-22)
+
+
+### Dependencies
+
+* Update dependency com.google.cloud:google-cloud-bigquery to v2.22.0 ([#1991](https://github.com/googleapis/java-bigquerystorage/issues/1991)) ([0684e7c](https://github.com/googleapis/java-bigquerystorage/commit/0684e7cf0f0b4d8ff343aee29f252489fe6c1ee7))
+* Update dependency com.google.cloud:google-cloud-shared-dependencies to v3.3.0 ([#2003](https://github.com/googleapis/java-bigquerystorage/issues/2003)) ([d146bec](https://github.com/googleapis/java-bigquerystorage/commit/d146becd5bf1022a4259c027a5e9b8db262170ea))
+
 ## [2.32.0](https://github.com/googleapis/java-bigquerystorage/compare/v2.31.1...v2.32.0) (2023-02-21)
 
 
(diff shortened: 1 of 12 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -196,6 +196,12 @@ class ConnectionWorker implements AutoCloseable {
    */
   private final String writerId = UUID.randomUUID().toString();
 
+  /*
+   * Test only exception behavior testing params.
+   */
+  private RuntimeException testOnlyRunTimeExceptionInAppendLoop;
```
yirutang, line 202: init this to null?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -383,6 +410,22 @@ private void maybeWaitForInflightQuota() {
     inflightWaitSec.set((System.currentTimeMillis() - start_time) / 1000);
   }
 
+  @VisibleForTesting
+  static long calculateSleepTime(long retryCount) {
+    return (long) Math.pow(5, retryCount);
```
yirutang, line 415: Are we backing off too much? Maybe base 2 with a max interval of say 1 min.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: added a max, but it's millisecond, do we still think 5 is too much?

yirutang: We have a total time out of retry 5 minutes. So we end up can only retry 7 times? As long as there is a max, I am OK.

GaoleMeng: Let's change to 2 to retry more times

yirutang (approved):

PR #1993, opened 2023-02-10
chore: add comment to clarify behavior setExecutorProvider

egreco12:
When creating a JsonStreamWriter builder, if the provided BigQueryWriteClient is not null, then any ExecutorProvider added by setExecutorProvider will not be used.  This is reasonable default behavior, but is not obvious.


yirutang (approved):

yirutang:
Let's close this PR, I am going to send out another fix to make it work better. Gaole and I verified that passing in a bigquery client never worked before and we still have a chance to make it right.

PR #2001, opened 2023-02-15
fix: Allow StreamWriter settings to override passed in BQ client setting

yirutang:
On StreamWriter, when user pass in a existing BQ client, we always used the old settings, and regional routing shouldn't have worked (since there is no header set).
From JsonWriter point of view, previously we always assign the internal BQ client to be null, so that the settings on Writer is respected and the BQ client passed into the JsonWriter is ignored (only used for JsonWriter level operations such as GetWriteStream). But the behavior is changed when allow BigQuery client to be passed through.

Clean up the behavior here to make it consistent and work for the long term. The behavior will be:
If bigquery client is passed in, then the setting will be inherited even though we always recreate a new client for the connection.
If bigquery client is passed in and StreamWriter has other explicit settings, we will override the setting use what is set explicitly on the Writer, and respect the rest of the bigquery client settings.


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -302,17 +302,37 @@ static boolean isDefaultStream(String streamName) {
...
+    if (builder.channelProvider != null) {
+      settingsBuilder.setTransportChannelProvider(builder.channelProvider);
+    }
+    if (builder.credentialsProvider != null) {
+      settingsBuilder.setCredentialsProvider(builder.credentialsProvider);
+    }
+    if (builder.executorProvider != null) {
+      settingsBuilder.setBackgroundExecutorProvider(builder.executorProvider);
+    }
+    if (builder.endpoint != null) {
```
GaoleMeng, line 331: this is always triggered, endpoint is never null, it's default to non-null and never assigned as null, so it will always overwrite the one given in the client:
https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L545
https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java#L597

GaoleMeng: missed the later part of this file.

GaoleMeng (approved):

PR #2017, opened 2023-02-24
fix: add client shutdown if request waiting in request queue for too long.

GaoleMeng:
this is needed as offline investigation, it's possible for grpc streaming lib to never trigger doneCallback / requestCallback under some scenario, e.g. when there's too many GAX threads. We should add this timeout to help connection spit out the requests on dead connections


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -659,6 +678,18 @@ private void appendLoop() {
     log.info("Append thread is done. Stream: " + streamName + " id: " + writerId);
   }
 
+  private void throwIfWaitCallbackTooLong() {
+    Duration milliSinceLastCallback =
+        Duration.between(lastRequestCallbackTriggerTime, Instant.now());
+    if (milliSinceLastCallback.compareTo(MAXIMUM_REQUEST_CALLBACK_WAIT_TIME) > 0) {
+      throw new IllegalStateException(
```
reuvenlax, line 685: I don't think this is really an IllegalStateException

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -659,6 +678,18 @@ private void appendLoop() {
     log.info("Append thread is done. Stream: " + streamName + " id: " + writerId);
   }
 
+  private void throwIfWaitCallbackTooLong() {
+    Duration milliSinceLastCallback =
+        Duration.between(lastRequestCallbackTriggerTime, Instant.now());
```
reuvenlax, line 683: If a StreamWriter is created and then there is a pause before using it, won't this trigger? (odd scenario, I know). Actually if a user ever pauses for 10 minutes, won't this cause the exception the be thrown the next time they use the StreamWriter?

GaoleMeng: The caller of the throwIfWaitCallbackTooLong is checking on the inflight queue size before triggering this check

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: Let's use a generic runtime exception, the goal is only to trigger uncaught exception handler

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
reuvenlax: Alternatively, would it make sense to store a timestamp in the RequestWrapper itself when it is added to the inflightRequestQueue? Since the queue is in order, you then can just check the timestamp at the head of the queue

reuvenlax: right, but append will add to this queue, right? So if append is called after a long delay, the appendLoop might hit this condition (if the appendLoop runs before the callback comes back).

GaoleMeng: Changed to a storing timestamp directly during creation of request wrapper

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -659,6 +672,17 @@ private void appendLoop() {
     log.info("Append thread is done. Stream: " + streamName + " id: " + writerId);
   }
 
+  private void throwIfWaitCallbackTooLong(Instant timeToCheck) {
+    Duration milliSinceLastCallback = Duration.between(timeToCheck, Instant.now());
+    if (milliSinceLastCallback.compareTo(MAXIMUM_REQUEST_CALLBACK_WAIT_TIME) > 0) {
+      throw new RuntimeException(
```
reuvenlax, line 678: Are you sure that a RuntimeException will be caught and will cancel things?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -911,11 +935,14 @@ static final class AppendRequestAndResponse {
...
     final StreamWriter streamWriter;
 
+    final Instant requestCreationTimeStamp;
+
     AppendRequestAndResponse(AppendRowsRequest message, StreamWriter streamWriter) {
       this.appendResult = SettableApiFuture.create();
       this.message = message;
       this.messageSize = message.getProtoRows().getSerializedSize();
       this.streamWriter = streamWriter;
+      this.requestCreationTimeStamp = Instant.now();
```
reuvenlax, line 945: Not sure if this is 100% right, since this gets called when the request is added to waitingRequestQueue. I think we want a timestamp that is set  close to when the actual RPC goes out.

GaoleMeng: I was trying to start counting time when user call Append. But let's change to set the timestamp when the request is added to the inflight queue.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: yeah: https://github.com/googleapis/java-bigquerystorage/blob/main/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java#L249

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
reuvenlax: The idea is to detect stuck RPCs. If things are stuck in the waitingRequestQueue (and nothing is stuck in the inflight queue), then presumably something else is very wrong.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -65,6 +66,14 @@ class ConnectionWorker implements AutoCloseable {
   // Maximum wait time on inflight quota before error out.
   private static long INFLIGHT_QUOTA_MAX_WAIT_TIME_MILLI = 300000;
 
+  /*
+   * Maximum time waiting for request callback before shutting down the connection.
+   *
+   * We will constantly checking how much time we have been waiting for the next request callback
+   * if we wait too much time we will start shutting down the connections and clean up the queues.
+   */
+  private static Duration MAXIMUM_REQUEST_CALLBACK_WAIT_TIME = Duration.ofMinutes(10);
```
yirutang, line 75: To keep it safe, let's put a longer time say 30 minutes.

GaoleMeng: Let's use 20 minutes, 30 minutes sound crazy long for a pipeline to stop

yirutang: Right now the user waits forever on the queue, I am just afraid some user does crazy things. I am fine with 20 min.

yirutang (approved):

PR #2037, opened 2023-03-13
fix: correct AppendSerializtionError typo

Neenu1995:
Since we don't want to break customers using the `AppendSerializtionError`(with typo) and at the same time, move all the reference we have to that class in the library, to the class without typo, it made sense to use class `AppendSerializationError extends AppendSerializtionError` (super class has typo).

The other way round would mean, we move the code in `AppendSerializtionError` to `AppendSerializationError` for us to use in the library. That would break the external customers, if we don't move the code, that will cause code duplication.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -244,6 +244,14 @@ public final class Exceptions {
     }
   }
 
+  public static class AppendSerializationError extends AppendSerializtionError{
+
+    public AppendSerializationError(int codeValue, String description, String streamName,
+        Map<Integer, String> rowIndexToErrorMessage) {
+      super(codeValue, description, streamName, rowIndexToErrorMessage);
+    }
+  }
+
   /** This exception is used internally to handle field level parsing errors. */
   public static class FieldParseError extends IllegalArgumentException {
     private final String fieldName;
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
@@ -244,9 +244,12 @@ public final class Exceptions {
     }
   }
 
-  public static class AppendSerializationError extends AppendSerializtionError{
+  public static class AppendSerializationError extends AppendSerializtionError {
 
-    public AppendSerializationError(int codeValue, String description, String streamName,
+    public AppendSerializationError(
+        int codeValue,
+        String description,
+        String streamName,
         Map<Integer, String> rowIndexToErrorMessage) {
       super(codeValue, description, streamName, rowIndexToErrorMessage);
     }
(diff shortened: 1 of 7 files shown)


Neenu1995:
cc: @suztomo 
This is to fix the typo without a breaking change.

suztomo:
"ci / clirr" passing is a nice.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -244,6 +244,17 @@ public String getStreamName() {
     }
   }
 
+  public static class AppendSerializationError extends AppendSerializtionError {
```
suztomo, line 247: Javadoc? (I expect a copy of AppendSerializtionError) Explanation of typo and backward compatibility is nice.

suztomo (commented):
I think there are 2 options: `class AppendSerializationError extends AppendSerializtionError` (super class has typo) or  `class AppendSerializtionError extends AppendSerializationError` (base class has typo).

This pull request seems to have chosen the former. Would you write down the rationale?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
Neenu1995: Since we don't want to break customers using the AppendSerializationError(with typo) and at the same time, move all the reference we have to that class in the library, to the class without typo, it made sense to use `class AppendSerializationError extends AppendSerializtionError` (super class has typo).

The other way round would mean, we move the code in `AppendSerializtionError` to `AppendSerializationError` for us to use in the library. That would break the external customers, if we don't move the code, that will cause code duplication.

Neenu1995:
@suztomo Would you mind taking another look?
Can you also confirm that, the class with the wrong typo name can be removed during next major version update?

suztomo:
Let's say you merge this. How would library users know they use `AppendSerializationError` (no typo)?

Neenu1995:
Customers who are not directly using this class will not be affected in any way.

Any customer who is directly using the AppendSerializtionError(with typo) will continue to use it but should eventually move to AppendSerializationError(without typo), if we remove the class with typo. 

The purpose of doing this work around right now is to introduce the class without typo, without introducing a breaking change now because we don't want a major version update now.

Eventually when we do a major version update, we should remove the class with typo. This will break customers who are directly using the class with typo. There is no way around it. For any users who may want to use the class from now, there is a warning on the class with typo that it will be removed. I could add a @Deprecated annotation as well.

farhan0102 (approved):

suztomo (approved):
I also think this is the best under the restriction of no breaking changes.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/Exceptions.java
```diff
@@ -216,10 +216,8 @@ public static StorageException toStorageException(Throwable exception) {
...
 
   /**
-   * This exception is thrown from {@link JsonStreamWriter#append()} when the client side Json to
-   * Proto serializtion fails. It can also be thrown by the server in case rows contains invalid
-   * data. The exception contains a Map of indexes of faulty rows and the corresponding error
-   * message.
+   * This class has a typo in the name. It will be removed soon. Please use {@link
+   * AppendSerializationError}
    */
   public static class AppendSerializtionError extends StatusRuntimeException {
```
yirutang, line 222: Need to update here?

yirutang: I see. I thought this is the final PR.

PR #2084, opened 2023-04-22
fix: Reduce timeout of waiting in queue from 15 minutes to 5 minutes and rephrase the log a bit 

GaoleMeng:
Since write api server side per request timeout is 2 minutes, it does not make sense to wait 15 minutes to determine whether a connection has been dead, reducing it to 3 minutes 

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -75,7 +75,7 @@ class ConnectionWorker implements AutoCloseable {
    * We will constantly checking how much time we have been waiting for the next request callback
    * if we wait too much time we will start shutting down the connections and clean up the queues.
    */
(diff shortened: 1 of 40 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -75,7 +75,7 @@ class ConnectionWorker implements AutoCloseable {
    * We will constantly checking how much time we have been waiting for the next request callback
    * if we wait too much time we will start shutting down the connections and clean up the queues.
    */
-  static Duration MAXIMUM_REQUEST_CALLBACK_WAIT_TIME = Duration.ofMinutes(15);
+  static Duration MAXIMUM_REQUEST_CALLBACK_WAIT_TIME = Duration.ofMinutes(3);
```
yirutang, line 78: Do we have retries? We could assume one api call costs 2 min. Let's make it 5 min to be safe.

GaoleMeng: Only retry will trigger is client side, let's allow two times retry, changed to 5 minutes

yirutang (approved):

PR #2085, opened 2023-04-22
refactor: Json to proto message refactoring

### LINKED ISSUE #2042: JsonToProtoMessage refactoring
I read through `JsonToProtoMessage` and would like refactor it as my first contribution. I'd like to know your opinion about the proposed changes and ask for some explanations.

1. All the fields should be final and CAPITALIZED (they are all deeply constant)
2. We should migrate from `import org.threeten.bp` to `java.time`. This would require migration of `com.google.cloud.bigquery.storage.v1.CivilTimeEncoder`. As we support Java 8, we should not need it anymore.
3. Code like `/*ignoreUnknownFields*/ false` should be replaced with `DONT_IGNORE_UNKNOWN_FIELDS` field
4. The `topLevel` parameter should be removed
5. `fillField` and `filRepeatedField` switch options should be extracted into methods. And even divided more to help JIT optimise the code.
6. The failing mechanism in `fillRepeatedField` should be replaced with throwing exception immediately. Otherwise one may think something can happen along the way to fix the situation. The call should be extracted into method
7. Two variants of `convertJsonToProtoMessage` method should be removed as they are used only in tests. The conversion should not support null in TableFieldSchema and FieldDescriptor. I think there is no case when it happens except tests. Is there? If yes, then it should be marked `@Nullable`
8. Comparing `"true"` with `toLowerCase()` should be replaced by `equalsIgnoreCase()`
9. The test should not use asterisks for imports
10. We should prefer `for(X x: xs)` over `for (int i = 0; i < xs.length; i++) {X x = xs[i];}`
11. The `JsonDataHasUnknownFieldException` should not be final and have protected constructor at the same time. It should be renamed to `DataHasUnknownFieldException` to be used by developers for other transformers.
12. Should we support conversion from JSON strings to booleans, ints, etc.?
13. Should we allow each element on the list to be of different type? Now it is allowed for some elements in the list to be `"true"` and some can be `true`.
14. For nested fields, there is no need to create additional message builder and ignoring it immediately. Is there?
15. There is no need to repeatedly check field descriptor type for repeated fields. The loop should be inside of each case. We can further optimise by checking the type of the first element on the list and assume all the files are of the same type.

Please tell me what you think.

arturowczarek:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-bigquerystorage/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Changes:
refactor: Use equalsIgnoreCase for JSON string boolean values in JsonToProtoMessage and use primitives for booleans
refactor: Use enhanced for in JsonToProtoMessage
refactor: Properly reference to static field INSTANCE in JsonToProtoMessage
refactor: Remove unnecessary LOG field from JsonToProtoMessage
refactor: Remove unnecessary casts from JsonToProtoMessage methods and remove usage of soon deprecated (Java 9) Long constructor
refactor: Remove unnecessary variables "message" from JsonToProtoMessage methods
refactor: Remove topLevel parameter from JsonToProtoMessage methods
refactor: Extract magic variables to constants
refactor: Capitalize deep constants in JsonToProtoMessage and make them final

Fixes #2042 ☕️

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
@@ -103,11 +103,9 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
           .toFormatter()
           .withZone(ZoneOffset.UTC);
   private final String ROOT_JSON_SCOPE = "root";
-  private final boolean IS_TOP_LEVEL = true;
-  private final boolean NOT_TOP_LEVEL = false;
   private final boolean DONT_ACCEPT_UNKNOWN_FIELDS = false;
 
-    /** You can use {@link JsonToProtoMessage.INSTANCE} instead */
+  /** You can use {@link JsonToProtoMessage.INSTANCE} instead */
   public JsonToProtoMessage() {}
 
   public static DynamicMessage convertJsonToProtoMessage(
@@ -156,7 +154,7 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
     Preconditions.checkNotNull(protoSchema, "Protobuf descriptor is null.");
     Preconditions.checkState(json.length() != 0, "JSONObject is empty.");
 
-    return convertToProtoMessage(protoSchema, null, json, ROOT_JSON_SCOPE, IS_TOP_LEVEL, DONT_ACCEPT_UNKNOWN_FIELDS);
+    return convertToProtoMessage(protoSchema, null, json, ROOT_JSON_SCOPE, DONT_ACCEPT_UNKNOWN_FIELDS);
   }
 
   /**
@@ -181,7 +179,6 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
         tableSchema.getFieldsList(),
         json,
         ROOT_JSON_SCOPE,
-        IS_TOP_LEVEL,
         DONT_ACCEPT_UNKNOWN_FIELDS);
   }
 
@@ -208,7 +205,6 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
         tableSchema.getFieldsList(),
         json,
         ROOT_JSON_SCOPE,
-        IS_TOP_LEVEL,
         ignoreUnknownFields);
   }
 
@@ -218,7 +214,6 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
    * @param protoSchema
    * @param json
    * @param jsonScope Debugging purposes
-   * @param topLevel checks if root level has any matching fields.
    * @throws IllegalArgumentException when JSON data is not compatible with proto descriptor.
    */
   private DynamicMessage convertToProtoMessage(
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -102,8 +99,10 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
           .optionalEnd()
           .toFormatter()
           .withZone(ZoneOffset.UTC);
+  private final String ROOT_JSON_SCOPE = "root";
+  private final boolean DONT_ACCEPT_UNKNOWN_FIELDS = false;
```
yirutang, line 103: It is a bit strange to have this constant. I think it is easier to read just to keep the bollean.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -25,13 +25,11 @@
 import com.google.protobuf.Descriptors.Descriptor;
```
yirutang, line 25: Thanks for improving the code!
Could you provide a list of things refactored in the PR description?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
arturowczarek: Are you sure? It explicitly names the intention. Otherwise you may mix boolean values:
`return convertToProtoMessage(protoSchema, null, json, "root", true, false);`
Can you tell what is `true` and `false` here? Your IDE will probably help you here, but what about GitHub review? Are you sure you won't put `topLevel` value in place of `ignoreUnknownArguments` parameter?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
arturowczarek: Updated

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
yirutang: Usually we will just annotate the parameters with the argument name. So it is just:
 protoSchema,
        tableSchema.getFieldsList(),
        json,
        "root",
        /*topLevel=*/ true,
        /*ignoreUnknownFields*/ false);

arturowczarek: Do you want me to inline only `DONT_ACCEPT_UNKNOWN_FIELDS` or `ROOT_JSON_SCOPE` too?
What IDE do you use? I thought all of them display the names of the parameters where constants are used. See the screenshot:
Screenshot 2023-04-27 at 19 30 11

yirutang: DONT_ACCEPT_UNKNOWN_FIELDS and ROOT_JSON_SCOPE. Thanks.

yirutang (approved):

Issue #2083, opened 2023-04-21
JsonStreamWriter.Builder#setEnableConnectionPool does not obey argument passed


bastewart:
#### Code example

`setEnableConnectionPool` does not obey the argument passed, instead however it is called connection pooling is enabled.

You can see this in the current master branch below (introduced - presumably by mistake! - in https://github.com/googleapis/java-bigquerystorage/pull/2048):

https://github.com/googleapis/java-bigquerystorage/blob/4fc828674fc63f398aafe822ae7d8532fb9e1a45/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java#L322-L325

Sorry, I have never contributed to this repo otherwise I'd just open a PR with a fix myself!

Neenu1995:
Thanks for opening the issue @bastewart.
Opened [PR](https://github.com/googleapis/java-bigquerystorage/pull/2096) to fix it.

Feel free to open a PR in the future, contributors are always welcome.

bastewart:
Thank you! 🙏

bastewart:
Fixed in https://github.com/googleapis/java-bigquerystorage/issues/2096. Thank you!

PR #2096, opened 2023-05-01
fix: pass the parameter value of enableConnectionPool instead of true always

Neenu1995:
Fixes #2085


yirutang (approved):

bastewart:
Just a note that this actually fixes https://github.com/googleapis/java-bigquerystorage/issues/2083! I'll close that issue myself 👍

PR #2097, opened 2023-05-01
feat: Improve client settings

yirutang:
1. Add ExecutorProvider to sample
2. Add KeepAlive setting to default set up.


samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -151,7 +153,10 @@ public void initialize(TableName parentTable)
       // For more information about JsonStreamWriter, see:
       // https://googleapis.dev/java/google-cloud-bigquerystorage/latest/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.html
       streamWriter =
-          JsonStreamWriter.newBuilder(parentTable.toString(), BigQueryWriteClient.create()).build();
+          JsonStreamWriter.newBuilder(parentTable.toString(), BigQueryWriteClient.create())
```
GaoleMeng, line 156: our example code create the client by "BigQueryWriteClient.create()" does that means we are not passing any keepalive setting?

yirutang: I added as a default setting. If I use a new transport in the sample, then the default setting will be overwritten.

I think in most cases, users doesn't need to reset the transport settings, our default is covering all that's needed (being set by Dataflow).

GaoleMeng (approved):

PR #2122, opened 2023-05-25
fix: bug fix for streamWriter & jsonStreamWriter

GaoleMeng:
1. When the user explicitly provides a table schema to JsonStreamWriter, whenever there is a schema update, we will overwrite the user provided table schema with the updated schema. And we will use this schema to set `writer_schema`. This will make the frontend not able to set default value, because default value is only set when the columns are not specified in the writer_schema.
2. Also we combined another fix for location string for multiplexing comparision, we shouldn't use '==' to compare 

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/SchemaAwareStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/SchemaAwareStreamWriter.java
@@ -1,0 +1,630 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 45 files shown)


yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/SchemaAwareStreamWriter.java
```diff
@@ -125,6 +130,9 @@ private Message buildMessage(T item)
       return this.toProtoConverter.convertToProtoMessage(
           this.descriptor, this.tableSchema, item, ignoreUnknownFields);
     } catch (Exceptions.DataHasUnknownFieldException ex) {
+      if (this.skipRefreshStreamWriter) {
```
yirutang, line 133: false?

yirutang: I think we should have unit test for this...

GaoleMeng: This is correct I think -- it means when refresh is skipped we directly return error
It's already covered by unit tests

PR #2140, opened 2023-06-12
chore: Update ResourceHeaderTest.java

Neenu1995:


Neenu1995:
The units 17 failure is likely the same issue as https://github.com/google/libphonenumber/pull/2877

PR #2197, opened 2023-07-20
feat: Support gRPC Compression

yifatgortler:
Enable gRPC compression when a compressor name is provided to JsonStreamWriter/StreamWriter.
Only `gzip` compression is supported at the moment.

Please note that while it will reduce the network usage, it will increase CPU usage. Please verify the trade off per use case.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
@@ -20,7 +20,6 @@ import com.google.api.gax.batching.FlowControlSettings;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.ExecutorProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
-import com.google.common.base.Preconditions;
 import com.google.protobuf.Descriptors;
 import java.io.IOException;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -336,6 +337,22 @@ public Builder setLocation(String location) {
...
+     * Sets the compression to use for the calls. The compressor must be of type gzip.
+     *
+     * @param compressorName
+     * @return Builder
+     */
+    public Builder setCompressorName(String compressorName) {
+      Preconditions.checkNotNull(compressorName);
+      Preconditions.checkArgument(
+          compressorName.equals("gzip"),
+          "Compression of type \"%s\" isn't supported, only \"gzip\" compression is supported.",
```
yirutang, line 350: Only check it at the bottom layer (ConnectionWorker) is enough?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/SchemaAwareStreamWriter.java
```diff
@@ -609,6 +616,22 @@ public Builder<T> setLocation(String location) {
...
 
+    /**
+     * Sets the compression to use for the calls. The compressor must be of type gzip.
+     *
+     * @param compressorName
+     * @return Builder
+     */
+    public Builder<T> setCompressorName(String compressorName) {
+      Preconditions.checkNotNull(compressorName);
+      Preconditions.checkArgument(
```
yirutang, line 627: Same as above.

yirutang (commented):
Thanks for getting this fixed!

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
yifatgortler: Removed from JsonStreamWriter and SchemaAwareStreamWriter, and left it in StreamWriter. Verified that it's what we usually do.

yirutang (approved):

PR #2212, opened 2023-07-27
fix(bigquery): Make exponential backoff retry second based

### LINKED ISSUE #2211: Current ConnectionWorker retry strategy is too aggressive
Thanks for stopping by to let us know something could be better!

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ..."): **BigQuery**
2. OS type and version: ALL
3. Java version: ALL
4.  version(s): Latest 2.41.0

#### Steps to reproduce

  1. Using Spark BigQuery connector
  2. Write a huge dataset to BigQuery
  3. Debug why it's super slow, the workers are waiting in `com.google.cloud.bigquery.storage.v1.ConnectionWorker#calculateSleepTimeMilli`

Looking at the code, the retry steps are too aggressive (1 milliseconds, 2 milliseconds, ..., 60 seconds), I suspect this is a bug and it was not intended to be this aggressive when retrying (the first 7 retries will happen in less than 256 milliseconds!)

mina-asham:

Fixes #2211 ☕️

- Currently these retries are doing retries in milliseconds, which is way too aggressive, the current retry steps are (1ms, 2ms, 4ms, 8ms, 16ms, 32ms, 64ms, 128ms, 256ms, 512ms, 1024ms, ...), it's highly likely the first 7 retries will fail with rate limiting specially in a big workload (specifically this is causing an issue in the Spark bigquery connector), I suspect the exponential retries here were meant to be in seconds (though even if it wasn't meant to, this might be a slightly better approach to reduce the load on the servers
- The new steps would be: 1s, 2s, 4s, 8s, 16s, 32s, 60s (repeated till we exhaust the 5 minutes max retries)

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -502,7 +502,7 @@ class ConnectionWorker implements AutoCloseable {
 
   @VisibleForTesting
   static long calculateSleepTimeMilli(long retryCount) {
-    return Math.min((long) Math.pow(2, retryCount) * 50, 60000);
+    return (long) Math.min(Math.pow(2, retryCount) * 50, 60000);
   }
 
   @VisibleForTesting
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -502,7 +502,7 @@ class ConnectionWorker implements AutoCloseable {
 
   @VisibleForTesting
   static long calculateSleepTimeMilli(long retryCount) {
-    return Math.min((long) Math.pow(2, retryCount) * 1000, 60000);
+    return Math.min((long) Math.pow(2, retryCount) * 50, 60000);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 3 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -502,7 +502,7 @@ private void maybeWaitForInflightQuota() {
 
   @VisibleForTesting
   static long calculateSleepTimeMilli(long retryCount) {
-    return Math.min((long) Math.pow(2, retryCount), 60000);
+    return Math.min((long) Math.pow(2, retryCount) * 1000, 60000);
```
GaoleMeng, line 505: Thanks for submitting this, the original intention for retrying reconnection at so frequent level is the majority of the connection reconnection does not need seconds of waiting. Were you seeing retry failure with small retry interval?

mina-asham: The problem I saw is probably more throttling rather than specific failures, this is used directly in the bigquery-spark-connector and the problem I am seeing is the connector slowing down to a failure writing data into bigquery, this is one of the two points I found led to this (there might be more, I might be mistaken too), Seconds maybe too much yes, bit retrying within a few millis is too aggressive I think, how about we do a base of 50ms or 100ms instead of 1s? That would get the best of both worlds, retry fast the first few times, and then back off exponentially

GaoleMeng: Yeah, I think 50ms makes sense to me. Let's use that value

mina-asham: Updated to 50ms

GaoleMeng (approved):

mina-asham:
@shollyman can we merge this one?

mina-asham:
@GaoleMeng what's the next steps here, this has been approved for almost 3 months now?

PR #2230, opened 2023-08-16
feat: expose settings to configure default missing value interpretation.

GaoleMeng:
Expose a new API to configure the default missing value interpretation.

This value will affect the column that has default value defined, but not presented in the missing value map.

For JsonStreamWriter:
1. If this value is set to `DEFAULT_VALUE`, we will always populate default value if the
field is missing from json and default value is defined in the column.
2. If this value is set to `NULL_VALUE`, we will always not populate default value.

For StreamWriter:
It will be the default configuration for the value not presented in the missing value map


yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonStreamWriter.java
```diff
@@ -346,6 +346,22 @@ public Builder setCompressorName(String compressorName) {
...
     }
 
+    /**
+     * Sets the default missing value interpretation value if the column is not presented in the
+     * missing_value_interpretations map.
+     *
+     * <p>If this value is set to `DEFAULT_VALUE`, we will always populate default value if the
+     * field is missing from json and default value is defined in the column.
+     *
+     * <p>If this value is set to `NULL_VALUE`, we will always not populate default value.
```
yirutang, line 356: Change the comment of 'NULL_VALUE' to: we will always populate NULL to the missing fields, which is currently our default behavior.

PR #2295, opened 2023-10-30
feat: add support for proto3 optional tag

agrawal-siddharth:
Fixes internal issue #304838994 whereby a NULL value was being inserted rather than the default value.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ProtoSchemaConverter.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ProtoSchemaConverter.java
@@ -102,6 +102,17 @@ public class ProtoSchemaConverter {
           enumTypes.add(enumFullName);
         }
       }
+      // The protobuf payload will be decoded as proto2 on the server side. The schema is also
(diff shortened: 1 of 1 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ProtoSchemaConverter.java
```diff
@@ -102,6 +102,12 @@ private static ProtoSchema convertInternal(
           enumTypes.add(enumFullName);
         }
       }
+      if (resultField.hasProto3Optional()) {
```
yirutang, line 105: nit: add some comment about why we need to clear the fields. Basically the binary string will be decoded as proto2 in the service side, and ProtoSchema needs to be proto2, so we need to clear proto3 only features. Proto2 and proto3 are binary compatible so it is OK.

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ProtoSchemaConverter.java
agrawal-siddharth: Done

PR #2319, opened 2023-11-14
feat: add new configuration for nightly retry-related tests.

egreco12:


.kokoro/build.sh
```diff
@@ -112,6 +112,28 @@ clirr)
     mvn -B -Denforcer.skip=true clirr:check
     RETURN_CODE=$?
     ;;
+retry_quota)
+    mvn -B ${INTEGRATION_TEST_ARGS} \
+      -Penable-integration-tests \
```
Neenu1995, line 117: Enabling this profile will run all the other IT tests also. Do we need that?

.kokoro/nightly/retry_non_quota.cfg
```diff
@@ -0,0 +1,11 @@
+# Format: //devtools/kokoro/config/proto/build.proto
```
Neenu1995, line 1: Also add some other env vars that the kokoro set up require.
[This](https://github.com/googleapis/java-bigquery/blob/main/.kokoro/nightly/nightly-integration.cfg) file is a good reference.

egreco12: Done; let me know if I missed something.  Additionally, is the test runner's SA java-it-service-account or it-service-account?  Currently only it-service-account@gcloud-devel.iam.gserviceaccount.com is allowlisted for the special projects.

.kokoro/build.sh
egreco12: We do not; I changed the command to test with explicit -Dtest args, let me know if that still needs work

.kokoro/nightly/retry_non_quota.cfg
Neenu1995: I believe, java-it-service-account is the name for the secret manager where the key is stored.

Neenu1995 (approved):

PR #2329, opened 2023-11-22
chore: add -U and -C to mvn install

Neenu1995:


suztomo:
Just curious; `-U --update-snapshots Forces a check for updated releases and snapshots on remote repositories` Did you encounter the need of this?

Neenu1995:
This PR addresses 2 issues. One is random build failures with package not found error. I suspect this is because of a broken dependency update(?).
I am trying out different variations to see which one works best.
-U seemed like a good option since it forces an update instead of using the cache.

PhongChuong (approved):

suztomo:
Thanks. If you see this change resolves problems that appear in other places. I’d like to apply it centrally.

PR #2437, opened 2024-03-07
feat: Add short, integer, long, boolean conversions into string

### LINKED ISSUE #2436: JsonStreamWriter simple types to string conversion
We are trying to migrate from legacy streaming API but we encountered following difference when using storage api with default stream.
Using storage writer, integer value in data is NOT automatically converted to string for fields with string type in schema. Oddly enough other way around, string in data but integer in schema, automatic conversion happens when string in data value actually is number string, for example "1"

Looking into JsonToProtoMessage code, lots of conversion is performed in different cases therefore I believe simple conversion like this should be done too.
If it is unacceptable for some reason, we can add boolean field (allowToStringConversions) to JsonStreamWriter Builder as feature flag to enable these conversions on request rather then by default.

Here is PR with suggested changes - https://github.com/googleapis/java-bigquerystorage/pull/2437

anthavio:
We are trying to migrate from legacy streaming API but we encountered following difference when using storage api with default stream.
Using storage writer, integer value in data is NOT automatically converted to string for fields with string type in schema. Oddly enough other way around, string in data but integer in schema, automatic conversion happens when string in data value actually is number string, for example "1"

Looking into JsonToProtoMessage code, lots of conversion is performed in different cases therefore I believe simple conversion like this should be done too. 
If it is unacceptable for some reason, we can add boolean field (allowToStringConversions) to JsonStreamWriter Builder as feature flag to enable these conversion on request rather then by default

Fixes #2436☕️

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If you are using Maven without the BOM, add this to your dependencies:
 If you are using Gradle 5.x or later, add this to your dependencies:
 
 ```Groovy
-implementation platform('com.google.cloud:libraries-bom:26.33.0')
+implementation platform('com.google.cloud:libraries-bom:26.34.0')
 
 implementation 'com.google.cloud:google-cloud-bigquerystorage'
 ```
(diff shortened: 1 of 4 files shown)


yirutang (approved):

yirutang:
lint error, please run:
mvn com.coveo:fmt-maven-plugin:format

PR #2445, opened 2024-03-09
fix: shutdown stream connection when timeout exception occurs

agrawal-siddharth:
Since the timeout exception results in the ConnectionWorker's appendLoop terminating, we must perform all shutdown steps that are performed there.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -353,7 +353,8 @@ class ConnectionWorker implements AutoCloseable {
           } finally {
             lock.unlock();
           }
-          cleanupInflightRequests();
+          cleanupConnectionAndRequests(
+              /* avoidBlocking= */ true); // don't perform blocking operations while on user thread
         });
     this.appendThread.start();
   }
@@ -812,7 +813,10 @@ class ConnectionWorker implements AutoCloseable {
         this.streamConnection.send(originalRequestBuilder.build());
       }
     }
+    cleanupConnectionAndRequests(/* avoidBlocking= */ false);
+  }
 
+  private void cleanupConnectionAndRequests(boolean avoidBlocking) {
     log.info(
         "Cleanup starts. Stream: "
             + streamName
@@ -828,7 +832,9 @@ class ConnectionWorker implements AutoCloseable {
     // We can close the stream connection and handle the remaining inflight requests.
     if (streamConnection != null) {
       this.streamConnection.close();
-      waitForDoneCallback(3, TimeUnit.MINUTES);
+      if (!avoidBlocking) {
+        waitForDoneCallback(3, TimeUnit.MINUTES);
+      }
     }
 
     // At this point, there cannot be more callback. It is safe to clean up all inflight requests.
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/it/ITBigQueryWriteManualClientTest.java
```diff
@@ -1612,4 +1615,70 @@ public void testLargeRequest() throws IOException, InterruptedException, Executi
       assertEquals("50", queryIter.next().get(0).getStringValue());
     }
   }
+
+  @Test
+  public void testTimeoutException() throws IOException, InterruptedException, ExecutionException {
+    try {
+      JsonStreamWriter.setMaxRequestCallbackWaitTime(
+          Duration.ofMillis(10)); // induce timeout exception
```
yirutang, line 1623: I am not entirely sure if tests will be executed in parallel, they usually will. If this is a static var, it could be problematic. I think it is safer to unit test this? In unit test, we can delay the callback for as long as we want, is the unit test not enough to hit the problem?

yirutang: Or we could introduce a local timeout var on the object, maybe call it SetLocalMaxRequestCallbackWaitTime or something. Maybe this value shouldn't be global to start with... But if we change the behavior now, it would be a breaking change.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -812,7 +812,10 @@ private void appendLoop() {
         this.streamConnection.send(originalRequestBuilder.build());
       }
     }
+    cleanup();
+  }
 
+  private void cleanup() {
```
yirutang, line 818: name this shutdownAndCleanup?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -812,7 +812,10 @@ private void appendLoop() {
         this.streamConnection.send(originalRequestBuilder.build());
       }
     }
+    cleanup();
```
yirutang, line 815: Why we need to call it here?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: I renamed it to cleanupConnectionAndRequests().

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: This cleanup function is shared by the code at the end of the appendLoop() and when an uncaught exception is thrown. In both cases the append thread needs to be terminated, and the inflight requests cleaned up.

yirutang (approved):

PR #2447, opened 2024-03-12
fix: add version id to the client lib header

yirutang:


agrawal-siddharth (approved):
Can you show what a sample traceId with the library version included looks like?

yirutang:
> Can you show what a sample traceId with the library version included looks like?

it should be like "java-streamwriter:3.0.1 user:trace" Same format as the managed go client lib.

> Can you show what a sample traceId with the library version included looks like?

PR #2499, opened 2024-05-11
fix: reset send timestamp each time a request is sent

agrawal-siddharth:
Each time a request is removed from the inflight queue its send timestamp should be removed by setting it to null as we are no longer waiting for a response to arrive for it.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -699,7 +699,10 @@ class ConnectionWorker implements AutoCloseable {
         hasMessageInWaitingQueue.await(100, TimeUnit.MILLISECONDS);
         // Check whether we should error out the current append loop.
         if (inflightRequestQueue.size() > 0) {
-          throwIfWaitCallbackTooLong(inflightRequestQueue.getFirst().requestCreationTimeStamp);
+          Instant sendInstant = inflightRequestQueue.getFirst().requestSendTimeStamp;
+          if (sendInstant != null) {
+            throwIfWaitCallbackTooLong(sendInstant);
+          }
         }
 
         // Copy the streamConnectionIsConnected guarded by lock to a local variable.
@@ -711,7 +714,9 @@ class ConnectionWorker implements AutoCloseable {
           // from inflightRequestQueue and prepent them onto the waitinRequestQueue. They need to be
           // prepended as they need to be sent before new requests.
           while (!inflightRequestQueue.isEmpty()) {
-            waitingRequestQueue.addFirst(inflightRequestQueue.pollLast());
+            AppendRequestAndResponse requestWrapper = inflightRequestQueue.pollLast();
+            requestWrapper.requestSendTimeStamp = null;
+            waitingRequestQueue.addFirst(requestWrapper);
           }
 
           // If any of the inflight messages were meant to be ignored during requestCallback, they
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/StreamWriterTest.java
```diff
@@ -1809,6 +1809,25 @@ public void testAppendSuccessAndInternalQuotaErrorRetrySuccess() throws Exceptio
...
+        com.google.rpc.Status.newBuilder().setCode(Code.RESOURCE_EXHAUSTED.ordinal()).build());
+    testBigQueryWrite.addStatusException(
+        com.google.rpc.Status.newBuilder().setCode(Code.RESOURCE_EXHAUSTED.ordinal()).build());
+    testBigQueryWrite.addStatusException(
+        com.google.rpc.Status.newBuilder().setCode(Code.RESOURCE_EXHAUSTED.ordinal()).build());
+    testBigQueryWrite.addResponse(createAppendResponse(0));
+
+    ApiFuture<AppendRowsResponse> appendFuture1 =
+        writer.append(createProtoRows(new String[] {"A"}));
+    assertEquals(0, appendFuture1.get().getAppendResult().getOffset().getValue());
```
yirutang, line 1827: How is this tested?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -1256,7 +1262,7 @@ static final class AppendRequestAndResponse {
 
     TimedAttemptSettings attemptSettings;
 
-    Instant requestCreationTimeStamp;
+    Instant requestSendTimeStamp;
```
yirutang, line 1265: Add some comment about this value, it is the timestamp where request hit the network.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -711,7 +714,9 @@ private void appendLoop() {
           // from inflightRequestQueue and prepent them onto the waitinRequestQueue. They need to be
           // prepended as they need to be sent before new requests.
           while (!inflightRequestQueue.isEmpty()) {
-            waitingRequestQueue.addFirst(inflightRequestQueue.pollLast());
+            AppendRequestAndResponse requestWrapper = inflightRequestQueue.pollLast();
+            requestWrapper.requestSendTimeStamp = null;
```
yirutang, line 718: I am wondering if you need to set this to null since it seems you will always set it before sending?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/StreamWriterTest.java
agrawal-siddharth: In order for the test to succeed, the given request must complete successfully even after all the retries. The fake server is configured to fail 3 times with a quota error. This means the client will perform retry with exponential backoff. The fake server injects 1 second of delay for each response. In addition, the exponential backoff injects a couple of seconds of delay. This yields an overall delay of about 5 seconds before the request succeeds. If the request send timestamp was being set only once, this would eventually exceed the 4 second timeout limit, and throw an exception. With the current behavior, the request send timestamp is reset each time a retry is performed, so we never exceed the 4 second timeout limit.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: Done.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: Whenever a request is removed from the inflight queue, this signals that we no longer expect to receive a response for this request. Hence we should not be checking this request for the timeout limit using a stale send timestamp. There could be a scenario where a request is placed back into the inflight queue, but must wait behind an earlier request that had failed and had a sequence number. We don't want to check it for a timeout limit until it has been sent again.

yirutang (approved):

PR #2509, opened 2024-05-22
chore: disable unit test that had problems due to static variable

agrawal-siddharth:


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/StreamWriterTest.java
```diff
@@ -1809,6 +1809,7 @@ public void testAppendSuccessAndInternalQuotaErrorRetrySuccess() throws Exceptio
     writer.close();
   }
 
+  /* temporarily disable test as static variable is interfering with other tests
```
yirutang, line 1812: could we just introduce the singleon first and fix this test? We could remove the original config as a breaking change later.

agrawal-siddharth: This test is failing and blocking others from making successful checkins. That's why I would like to disable it now and make a proper fix as a followup.

yirutang (approved):

PR #2515, opened 2024-05-28
feat: enable hermetic library generation

diegomarquezp:
This is part of the effort to enable multi-version support in hand-written libraries. More context and details in [this doc](https://docs.google.com/document/d/1wrpyBtphdenM3BNelcnpBKGADYrGJUo686HXvSA0h-0/edit?pli=1&tab=t.0#heading=h.lbr9t8ls3oip).

This PR adds two workflows and their scripts:
 * `update_generation_config` which updates `generation_config.yaml` (added in this PR) to contain the latest `googleapis_committish` (source of truth of library generation) as well as other dependencies such as `gapic_generator_version`. This job runs nightly and creates a PR that modifies the config yaml.
 * `hermetic_library_generation` which reads `generation_config.yaml` and pushes a new commit to any PR whose latest commit affects `generation_config.yaml` (e.g. the nightly PR created by `update_generation_config`). It essentially translates changes in the config yaml to changes in the source code (of the generated part). This workflow will be a required check.

As an example of how they work in practice:
 1. Once per night, `update_generation_config` creates a PR that updates `generation_config.yaml` with the latest dependencies and googleapis committish
 2. `hermetic_library_generation` is triggered by this PR and generates the corresponding code changes. A new commit with the code changes is added to the PR. This step may occur for any PR affecting `generation_config.yaml`.

### Follow ups
 - [x] Add `CLOUD_JAVA_BOT_TOKEN` to repository for the hermetic generation workflow to work


generation_config.yaml
```diff
@@ -0,0 +1,42 @@
+gapic_generator_version: 2.41.0
+protoc_version: '25.3'
```
blakeli0, line 2: I don't think we need `protoc_version` and `template_excludes` anymore.

generation_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - .kokoro/presubmit/graalvm-native.cfg
+    - .kokoro/dependencies.sh
+    - .github/workflows/approve-readme.yaml
+    - .github/workflows/auto-release.yaml
+    - .github/workflows/ci.yaml
+    - .github/workflows/samples.yaml
+    - .kokoro/requirements.in
+    - .kokoro/requirements.txt
+libraries:
+    - api_shortname: "bigquerystorage"
```
blakeli0, line 22: I don't think we have the values wrapped with double quotes in google-cloud-java, e.g. [asset](https://github.com/googleapis/google-cloud-java/blob/main/generation_config.yaml#L228-L244). Can we keep them consistent with google-cloud-java?

generation_config.yaml
diegomarquezp: Done for all repos.

generation_config.yaml
diegomarquezp: Done for all repos.

PhongChuong (approved):

blakeli0 (approved):

PR #2525, opened 2024-06-07
chore: demonstrate how to set user-agent header

agrawal-siddharth:


samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStream.java
```diff
@@ -22,6 +22,8 @@
 import com.google.api.core.ApiFutures;
 import com.google.api.gax.core.FixedExecutorProvider;
 import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.rpc.FixedHeaderProvider;
```
yirutang, line 25: Setting useragent is not a common practice. Let's keep the e2e test instead.

agrawal-siddharth: I have moved the sample code that demonstrates how to set the UserAgent to an existing e2e test, namely ITBigQueryWriteManualClientTest.java.

yirutang (approved):

PR #2547, opened 2024-07-02
chore: fix api_description in generation_config

diegomarquezp:


proto-google-cloud-bigquerystorage-v1/src/main/proto/google/cloud/bigquery/storage/v1/arrow.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2023 Google LLC
```
JoeWang1127, line 1: Do you know why the copyright year changed?

diegomarquezp: This is a regression due to only changing the api description without updating the googleapis_committish. I'll close it and use the generated PR

PR #2555, opened 2024-07-10
feat: add profiler for request execution details for write api connection worker 

GaoleMeng:
Add a profiler to generate periodical report. 

In the next PR, we would start to add interested point at which we want to examine the performance.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
@@ -1,0 +1,269 @@
+package com.google.cloud.bigquery.storage.v1;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.PriorityQueue;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
+/**
+ * A profiler that would periodically generate a report for the past period with the latency report for the slowest
+ * requests. This is used for debugging only.
+ *
+ * The report will contain the execution details of the TOP_K slowest requests, one example:
+ * ```
+ * INFO: At system time 1720566109971, in total 2 finished during the last 60000 milliseconds, the top 10 long latency requests details report:
+ * -----------------------------
+ * 	Request uuid: request_1 with total time 1000 milliseconds
+ * 		Operation name json_to_proto_conversion starts at: 1720566109971, ends at: 1720566109971, total time: 200 milliseconds
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 800 milliseconds
+ * -----------------------------
+ * 	Request uuid: request_2 with total time 500 milliseconds
+ * 		Operation name json_to_proto_conversion starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * ```
+ */
+public class RequestProfiler {
+    enum OperationName {
+        // The total end to end latency for a request.
+        TOTAL_REQUEST("total_request_time"),
+        // Json to proto conversion time.
+        JSON_TO_PROTO_CONVERSION("json_to_proto_conversion"),
+        // Time spent to fetch the table schema when user didn't provide it.
+        SCHEMA_FECTCHING("schema_fetching"),
+        // Time spent within wait queue before it get picked up.
+        WAIT_QUEUE("wait_queue"),
+        // Time spent within backend to process the request.
+        BACKEND_LATENCY("backend_latency");
+        private final String operationName;
+
+        OperationName(String operationName) {
+            this.operationName = operationName;
+        }
+    }
+
+    private static final Logger log = Logger.getLogger(RequestProfiler.class.getName());
+
+    // Singleton for easier access.
+    public static final RequestProfiler REQUEST_PROFILER_SINGLETON = new RequestProfiler();
+
+    // Tunable static variable indicate how many top longest latency requests we should consider.
+    private static int TOP_K = 10;
+
+    // Tunable static variable indicate how often the report should be generated.
+    private static Duration FLUSH_PERIOD = Duration.ofMinutes(1);
+
+    // From request uuid to the profiler of individual request. This will be cleaned up periodically.
+    private final Map<String, IndividualRequestProfiler> idToIndividualOperation =
+            new ConcurrentHashMap<>();
+
+    private Thread flushThread;
+
+    // Mark an operation for a given request id to be start.
+    void startOperation(OperationName operationName, String requestUniqueId) {
+        idToIndividualOperation.putIfAbsent(requestUniqueId,
+                new IndividualRequestProfiler(requestUniqueId));
+        idToIndividualOperation.get(requestUniqueId).startOperation(operationName);
+    }
+
+    // Mark an operation for a given request id to be end.
+    void endOperation(OperationName operationName, String requestUniqueId) {
+        idToIndividualOperation.get(requestUniqueId).endOperation(operationName);
+    }
+
+    void flushReport() {
+        log.info(flushAndGenerateReportText());
+    }
+
+    void startPeriodicalReportFlushing() {
+        this.flushThread =
+                new Thread(
+                        new Runnable() {
+                            @Override
+                            public void run() {
+                                while (true) {
+                                    try {
+                                        TimeUnit.MILLISECONDS.sleep(FLUSH_PERIOD.toMillis());
+                                    } catch (InterruptedException e) {
+                                        log.warning("Flush report thread is interrupted by " + e.toString());
+                                        throw new RuntimeException(e);
+                                    }
+                                    flushReport();
+                                }
+                            }
+                        });
+        this.flushThread.start();
+    }
+
+    String flushAndGenerateReportText() {
+        RequestProfilerComparator comparator = new RequestProfilerComparator();
+
+        // Find the top k requests with the longest latency.
+        PriorityQueue<IndividualRequestProfiler> minHeap =
+                new PriorityQueue<IndividualRequestProfiler>(comparator);
+        Iterator<Entry<String, IndividualRequestProfiler>> iterator =
+                idToIndividualOperation.entrySet().iterator();
+        int finishedRequestCount = 0;
+        while (iterator.hasNext()) {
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,327 @@
...
+    private final String operationName;
+
+    OperationName(String operationName) {
+      this.operationName = operationName;
+    }
+  }
+
+  private static final Logger log = Logger.getLogger(RequestProfiler.class.getName());
+
+  // Control per how many requests we log one time for a dropped operation.
```
agrawal-siddharth, line 73: I'm not clear what is meant by "dropped"? Does this mean a request was rejected?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,327 @@
...
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 800 milliseconds
+ * -----------------------------
+ * 	Request uuid: request_2 with total time 500 milliseconds
+ * 		Operation name json_to_proto_conversion starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * ...
+ * </pre>
+ */
+public class RequestProfiler {
+  enum OperationName {
```
agrawal-siddharth, line 53: Why does the name of the operation need to be pre-defined or known in advance to the profiler? Is this just to ensure consistency in how it is used?

agrawal-siddharth (changes requested):
So if I understand correctly, once this profiler is available, subsequent commits will then start using it in various places within the codebase. There is definitely overlap with the metrics being generated for OpenTelemetry. Perhaps eventually we could automatically profile those metrics with this tool.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: When we cache too many request I would drop that, added an comment.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: Yeah consistency, so that we would deal with random name

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,329 @@
...
+      this.operationName = operationName;
+    }
+  }
+
+  private static final Logger log = Logger.getLogger(RequestProfiler.class.getName());
+
+  // Control per how many requests we log one time for a dropped operation.
+  // An operation can be dropped if we are caching too many requests (by default 100000) in memory. If we are at that
+  // state, any new operation for new requests would be dropped.
+  private static final int LOG_PER_DROPPED_OPERATION = 50;
```
agrawal-siddharth, line 76: OK. Now I understand this constant better. It really just helps with reducing the volume of log messages generated around dropped operations. Maybe it would be more clear if you defined this constant after defining the MAX_CACHED_REQUEST. Perhaps a better name would be DROPPED_OPERATION_LOG_FREQUENCY.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,329 @@
...
+ * </pre>
+ */
+public class RequestProfiler {
+  enum OperationName {
+    // The total end to end latency for a request.
+    TOTAL_REQUEST("total_request_time"),
+    // Json to proto conversion time.
+    JSON_TO_PROTO_CONVERSION("json_to_proto_conversion"),
+    // Time spent to fetch the table schema when user didn't provide it.
+    SCHEMA_FECTCHING("schema_fetching"),
```
agrawal-siddharth, line 59: typo here, should be SCHEMA_FETCHING

agrawal-siddharth (changes requested):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: I think, let's just remove this variable and related the logic, the upper cache bound is really large, I would say safe guarding such extreme case does not have too much value

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
agrawal-siddharth: Let's ensure there is consistency between the metric names identified here, and those currently used and future ones to be added for OpenTelemetry metrics. The latter are defined in this document: http://go/writeapi-telemetry. See the names used under "List of instruments".

If there is a need for profiling metrics there is likely a need for the same OpenTelemetry metric.

agrawal-siddharth (changes requested):
I'm wondering if tracking every single request will result in excessive memory consumption. To deal with this, you may need to:

(a) use histogram buckets to break down individual values into aggregate values, or

(b) dynamically keep track of only the K longest latency requests per type.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: I think the current overlap is the backend latency, let's change it to NETWORK_RESPONSE_LATENCY

GaoleMeng (commented):
Discussed offline that it's not easy to come up with a clean solution that would only keep the top k requests all the time to save memory. Since it's a debugging only class and each flush would release the memory of the past requests, we think it's fine to continue with the current solution

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,325 @@
...
+  // Mark an operation for a given request id to be end.
+  void endOperation(OperationName operationName, String requestUniqueId) {
+    if (!idToIndividualOperation.containsKey(requestUniqueId)) {
+      log.warning(
+          String.format(
+              "endOperation is triggered for request_id: %s that's hasn't "
+                  + "seen before, this is possible when "
+                  + "we are recording too much ongoing requests. So far we has dropped %s operations.",
+              requestUniqueId, droppedOperationCount));
+      droppedOperationCount.incrementAndGet();
```
agrawal-siddharth, line 122: Should this be incremented here as well? What if the calling code always calls startOperation() and endOperation()?

agrawal-siddharth (changes requested):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
GaoleMeng: Done

agrawal-siddharth (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,327 @@
...
+ * for the slowest requests. This is used for debugging only.
+ *
+ * <pre>
+ * The report will contain the execution details of the TOP_K slowest requests, one example:
+ *
+ * INFO: At system time 1720566109971, in total 2 finished during the last 60000 milliseconds, the top 10 long latency requests details report:
+ * -----------------------------
+ * 	Request uuid: request_1 with total time 1000 milliseconds
+ * 		Operation name json_to_proto_conversion starts at: 1720566109971, ends at: 1720566109971, total time: 200 milliseconds
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 800 milliseconds
```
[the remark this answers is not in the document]
GaoleMeng: The current way we can join is through timestamp, unless this request id can be passed to backend. 

Let's call it response_latency to make it clear

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,327 @@
...
+ * -----------------------------
+ * 	Request uuid: request_2 with total time 500 milliseconds
+ * 		Operation name json_to_proto_conversion starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * 		Operation name backend_latency starts at: 1720566109971, ends at: 1720566109971, total time: 250 milliseconds
+ * ...
+ * </pre>
+ */
+public class RequestProfiler {
```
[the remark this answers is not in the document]
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
```diff
@@ -0,0 +1,327 @@
...
+  enum OperationName {
+    // The total end to end latency for a request.
+    TOTAL_REQUEST("total_request_time"),
+    // Json to proto conversion time.
+    JSON_TO_PROTO_CONVERSION("json_to_proto_conversion"),
+    // Time spent to fetch the table schema when user didn't provide it.
+    SCHEMA_FECTCHING("schema_fetching"),
+    // Time spent within wait queue before it get picked up.
+    WAIT_QUEUE("wait_queue"),
```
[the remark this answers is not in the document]
GaoleMeng: We could, but that requires really hard interface change, I wanna shrink the scope to time related profiling for now.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/RequestProfiler.java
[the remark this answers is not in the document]
yirutang: I don't see it is changed? It is still a public class.

PR #2580, opened 2024-07-26
[test] protobuf to 3.25.4 for v2.47.0

zhumin8:
This is for lts6.x patch test only.


zhumin8:
failing test-samples-java-bigquerystorage because it does not exists
File .cloudbuild/samples_build.yaml not found

PR #2581, opened 2024-07-26
chore: add plumbing to allow customer to enable reporting OT metrics

agrawal-siddharth:
This change prepares for a future change that will actually consume the flag to enable/disable reporting of OpenTelemetry metrics.


yirutang (commented):
Is this flag actually used?

agrawal-siddharth:
No, as yet the flag is not used. Next checkin factor out the metrics reporting logic and will begin using it.

yirutang (approved):

PR #2615, opened 2024-08-21
feat: v1alpha client generation

PhongChuong:


blakeli0 (approved):

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -24,7 +24,7 @@ env:
 jobs:
   library_generation:
     # skip pull requests coming from a forked repository    
-    if: github.env.REPO_FULL_NAME == github.repository
```
diegomarquezp, line 27: @blakeli0 inlining the `org/repo` as an env var seems to break this `if` statement

blakeli0: Probably because `REPO_FULL_NAME` is not referenced as a variable like `$REPO_FULL_NAME`?

blakeli0 (commented):
@PhongChuong Can we update the title to be a `feat` since the generated changes are also part of the PR?

Linchin (approved):

PR #2699, opened 2024-10-08
fix: add stream name to every request when connection is created during multiplexing

GaoleMeng:
fix: this is not required, but we may want to add this as extra safety to make sure every request is landed onto the correct table.


google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
```diff
@@ -279,26 +275,20 @@ public void testAppendInSameStream_switchSchema() throws Exception {
 
         // We will get the request as the pattern of:
         // (writer_stream: t1, schema: schema1)
-        // (writer_stream: _, schema: _)
+        // (writer_stream: t1, schema: _)
```
yirutang, line 278: How do we know we are multiplexed before the actual switch?

GaoleMeng: the connection would be created via connectionPool

yirutang (approved):

PR #2780, opened 2024-11-20
feat: introduce `java.time` methods and variables

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).


diegomarquezp:
@lqiu96 [This check](https://github.com/googleapis/java-bigquerystorage/actions/runs/11942130315/job/33288467778?pr=2780) is failing with

```
Error:    JsonToProtoMessageTest.testTimestampRepeated:890 expected:<test_string_repeated: 10
test_string_t_z_repeated: 1648493279010000
test_long_repeated: 1687984085000000
test_int_repeated: 153480695
test_float_repeated: [15346](https://github.com/googleapis/java-bigquerystorage/actions/runs/11942130315/job/33288467778?pr=2780#step:6:15347)8069500
test_offset_repeated: 1649135171000000
test_timezone_repeated: 1649174771000000
test_saformat_repeated: 1534680660000000
> but was:<test_string_repeated: 10
test_string_t_z_repeated: 1648493279010000
test_long_repeated: 1687984085000000
test_int_repeated: [15348](https://github.com/googleapis/java-bigquerystorage/actions/runs/11942130315/job/33288467778?pr=2780#step:6:15349)0695
test_float_repeated: 153468069500
test_offset_repeated: 1649149571000000
test_timezone_repeated: 1649174771000000
test_saformat_repeated: 1534680660000000
>
```

The problem is with `test_offset_repeated`; it's the same as https://github.com/googleapis/sdk-platform-java/pull/3330#discussion_r1828424787, but the difference here is that we use `appendOffset()` with a default zone string of `+00:00`, while the one in `appendZoneOrOffsetId()` is  `Z`

https://github.com/googleapis/java-bigquerystorage/blob/5f1b30a94767a58bd3bbd509cccd0d5f34260704/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java#L95-L102

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/JsonToProtoMessage.java
```diff
@@ -93,14 +93,11 @@ public class JsonToProtoMessage implements ToProtoConverter<Object> {
...
+          .appendZoneOrOffsetId()
           .optionalEnd()
           .optionalStart()
           .appendZoneText(TextStyle.SHORT)
           .optionalEnd()
-          .optionalStart()
-          .appendLiteral('Z')
-          .optionalEnd()
           .toFormatter()
           .withZone(ZoneOffset.UTC);
```
diegomarquezp, line 102: This fix works for the current tests. Note that the literal `Z` is dropped from the config because `appendZoneOrOffsetId()` uses `Z` as the no-zone string

diegomarquezp: Added a test for `+00:00` in https://github.com/googleapis/java-bigquerystorage/pull/2780/commits/260c835c175dae427b9d518818511a288ec7ad21

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/util/TimeConversionUtils.java
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigquery.storage.util;
+
+public class TimeConversionUtils {
```
lqiu96, line 19: Why is this added in BQStorage and not used from sdk-platform-java?

google-cloud-bigquerystorage/clirr-ignored-differences.xml
```diff
@@ -212,5 +212,21 @@
...
     <method>void setMissingValueInterpretationMap(java.util.Map)</method>
   </difference>
+  <difference>
+    <!--This class is marked as @InternalApi "public for technical reasons"-->
+    <differenceType>6004</differenceType>
+    <className>com/google/cloud/bigquery/storage/*/stub/readrows/ApiResultRetryAlgorithm</className>
+    <field>DEADLINE_SLEEP_DURATION</field>
+    <from>org.threeten.bp.Duration</from>
+    <to>java.time.Duration</to>
+  </difference>
```
lqiu96, line 222: I don't think I see this in the PR. Is this an old implementation?

google-cloud-bigquerystorage/clirr-ignored-differences.xml
```diff
@@ -212,5 +212,21 @@
...
+  <difference>
+    <!--The retryDelay field is used by ApiResultRetryAlgorithm, which is marked as @InternalApi-->
```
lqiu96, line 224: Errors$IsRetryableStatusResult is still marked as public even if it's only used by `ApiResultRetryAlgorithm` in the BQStorage SDK. I think it's still accessible which is problematic.

google-cloud-bigquerystorage/src/test/proto/jsonTest.proto
```diff
@@ -152,8 +152,9 @@ message TestTimestamp {
   optional int64 test_int = 4;
   optional int64 test_float = 5;
   optional int64 test_offset = 6;
-  optional int64 test_timezone = 7;
-  optional int64 test_saformat = 8;
+  optional int64 test_zero_offset = 7;
```
lqiu96, line 155: To confirm: This was added to help with the offset test, right?

diegomarquezp: Yes, it's meant to confirm that valid offsets such as `+12:34` or "no offset" strings such as `Z` and `+00:00` are treated correctly.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/util/TimeConversionUtils.java
diegomarquezp: They are functions only used in BQS so far (confirmed other repos don't need them). However, the reason is simply because it's faster to implement here than waiting for a new release of gax with the utilities we have here. What if we follow up in gax and temporarily have these functions in this repo?

google-cloud-bigquerystorage/clirr-ignored-differences.xml
diegomarquezp: It doesn't show up as a direct change because it's declared as `Duration DEADLINE_SLEEP_DURATION`. The change in the file was done in the [import statement](https://github.com/googleapis/java-bigquerystorage/pull/2780/files#diff-6fed86adbffd5ca2037793806c832765de4b7b04dbcc85b325b93d61b9f3c33f) so the type of this internal var still changes.

google-cloud-bigquerystorage/clirr-ignored-differences.xml
diegomarquezp: @PhongChuong what's the rationale of `Errors$IsRetryableStatusResult` being public? Is it similar to the case of `ApiResultRetryAlgorithm` "visible for technical reasons"?
Also, does it make sense to change their datatypes directly, without `threeten` alternatives as done currently?

PhongChuong (commented):
Overall the changes looks good. Thank you for the PR, it looks like a lot of work tracking down all the changes needed.

google-cloud-bigquerystorage/clirr-ignored-differences.xml
yirutang: I am fine making it private. I don't think external user should use it.

PhongChuong (approved):

google-cloud-bigquerystorage/clirr-ignored-differences.xml
lqiu96: Ok, I see the class is marked with InternalApi. I think this is fine

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/util/TimeConversionUtils.java
lqiu96: > What if we follow up in gax and temporarily have these functions in this repo?

Sure. Can we make this class package-private until we migrate this to Gax? Can you also create an issue in sdk-platform-java for this?

lqiu96 (approved):
LGTM. Added a small comment if you could address

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/util/TimeConversionUtils.java
diegomarquezp: As discussed internally, left as `@InternalApi`, referencing the new issue in sdk-platform: https://github.com/googleapis/sdk-platform-java/issues/3412

PR #2789, opened 2024-12-05
feat: add support for byte and short in JSONStreamWriter

agrawal-siddharth:
Now customers can provide JSON input in byte or short format for a INT64 table column.


yirutang (approved):

yirutang:
Thanks for fixing this, please let the customer know about the release containing this fix!

PR #2852, opened 2025-02-06
fix: allow trace id to switch within the same connection

GaoleMeng:
This change will allow the trace id to be changeable within the same connection. So that the same connection pool could be used for multiple different trace id from different stream writer. We will pass a new trace id to server side once we detect a table switch

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -868,9 +868,7 @@ class ConnectionWorker implements AutoCloseable {
           // If we are at the first request for every table switch, including the first request in
           // the connection, we will attach both stream name and table schema to the request.
           destinationSet.add(streamName);
-          if (this.traceId != null) {
-            originalRequestBuilder.setTraceId(this.traceId);
-          }
+          originalRequestBuilder.setTraceId(wrapper.streamWriter.getFullTraceId());
         } else if (!isMultiplexing) {
           // If we are not in multiplexing and not in the first request, clear the stream name.
           originalRequestBuilder.clearWriteStream();
--- a/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
+++ b/google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
@@ -86,23 +86,50 @@ public class ConnectionWorkerTest {
   }
 
   @Test
-  public void testMultiplexedAppendSuccess() throws Exception {
+  public void testMultiplexedAppendSuccess_NonNullTraceId() throws Exception {
+    testMultiplexedIngestion(
+        /*sw1TraceId=*/ "header_1:trailer_1",
+        /*sw2TraceId=*/ "header_2:trailer_2",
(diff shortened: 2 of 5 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -1372,16 +1375,21 @@ static final class AppendRequestAndResponse {
     // If a response is no longer expected this is set back to null.
     Instant requestSendTimeStamp;
 
+    // Mark where the traffic coming from. Used for internal logging purpose.
+    String fullTraceId;
```
agrawal-siddharth, line 1379: Why not retrieve the traceId from streamWriter.getFullTraceId()?

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
```diff
@@ -97,11 +97,13 @@ public void testMultiplexedAppendSuccess() throws Exception {
           StreamWriter.newBuilder(TEST_STREAM_1, client)
               .setWriterSchema(createProtoSchema("foo"))
               .setLocation("us")
+              .setTraceId("some_head:trace_id_1")
               .build();
       StreamWriter sw2 =
           StreamWriter.newBuilder(TEST_STREAM_2, client)
               .setWriterSchema(createProtoSchema("complicate"))
               .setLocation("us")
+              .setTraceId("some_head:trace_id_2")
```
agrawal-siddharth, line 106: Do you also have a test where the StreamWriter traceId is left as NULL?

agrawal-siddharth (commented):
Although this is OK for now, eventually we will also want to update the traceId on the telemetryMetrics when we have a table switch. Since currently we only use traceId for Dataflow in the telemetryMetrics, and Dataflow does not attempt to change the traceId across streams, we don't need this now.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
GaoleMeng: ConnectionWorker does not have reference to individual stream writer

GaoleMeng: Since multiple SW can share the same connecctionWorker

google-cloud-bigquerystorage/src/test/java/com/google/cloud/bigquery/storage/v1/ConnectionWorkerTest.java
GaoleMeng: Done

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: I meant the streamWriter found in AppendRequestAndResponse. This is available for each request, right?

GaoleMeng: You are right, done

agrawal-siddharth (approved):

PR #2942, opened 2025-04-19
fix: don't start a retry timer if connection was closed due to being …

agrawal-siddharth:
…idle

An idle connection will be closed by the server after 10 minutes of inactivity. At some later time, the client may wish to resume sending traffic. At this time, in case we receive another disconnect event from the server, we should not count the time that has elapsed from the earlier idle disconnect. Instead, the time should begin from the latest disconnect event.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
@@ -1273,23 +1273,28 @@ class ConnectionWorker implements AutoCloseable {
             + writerId
             + " Final status: "
             + finalStatus.toString());
+    boolean closedIdleConnection =
+        finalStatus.toString().contains("Closing the stream because it has been inactive");
     this.lock.lock();
     try {
       this.streamConnectionIsConnected = false;
       this.telemetryMetrics.recordConnectionEnd(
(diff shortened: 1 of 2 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
```diff
@@ -1273,31 +1273,38 @@ private void doneCallback(Throwable finalStatus) {
...
+    boolean closedIdleConnection =
+        finalStatus.toString().contains("Closing the stream because it has been inactive");
     this.lock.lock();
     try {
       this.streamConnectionIsConnected = false;
       this.telemetryMetrics.recordConnectionEnd(
           Code.values()[Status.fromThrowable(finalStatus).getCode().ordinal()].toString());
       if (connectionFinalStatus == null) {
-        if (connectionRetryStartTime == 0) {
+        if (!closedIdleConnection && (connectionRetryStartTime == 0)) {
```
yirutang, line 1284: nit, I don't think you need the ()

yirutang (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/ConnectionWorker.java
agrawal-siddharth: Done

PR #2945, opened 2025-04-23
chore: add sample demonstrating sending a nested proto

agrawal-siddharth:


yirutang (approved):

samples/install-without-bom/pom.xml
```diff
@@ -18,6 +18,7 @@
   </parent>
 
   <properties>
+    <project.protobuf-java.version>3.25.4</project.protobuf-java.version>
```
PhongChuong, line 21: nit. Any particular reason to use 3.25.4 instead of a later 3.25.x?

PhongChuong (approved):

samples/install-without-bom/pom.xml
agrawal-siddharth: No. I simply matched the version listed in java-bigquerystorage/pom.xml. Should I change it to use a more recent version?

PhongChuong: Matching sounds good. IIRC, they get automatically updated periodically.

PR #2983, opened 2025-05-23
feat(bigquery): Integrate Otel Tracing in storage lib

whuffman36:
This PR adds OpenTelemetry tracing support to the BigQueryReadClient and the BigQueryEnhancedReadStub. Tracing is toggled with a boolean flag and is enabled through the BigQueryReadSettings when creating the BigQueryReadClient.

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/BigQueryReadClient.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/BigQueryReadClient.java
@@ -20,9 +20,9 @@ import com.google.api.gax.core.BackgroundResource;
 import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.bigquery.storage.v1.stub.EnhancedBigQueryReadStub;
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Scope;
-import io.opentelemetry.api.common.Attributes;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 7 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/BigQueryReadClient.java
```diff
@@ -287,7 +332,22 @@ public final UnaryCallable<CreateReadSessionRequest, ReadSession> createReadSess
    * </code></pre>
    */
   public final ServerStreamingCallable<ReadRowsRequest, ReadRowsResponse> readRowsCallable() {
-    return stub.readRowsCallable();
+    Span readRowsCallable = null;
+    if (settings.isOpenTelemetryEnabled()) {
+      readRowsCallable =
+          settings
+              .getOpenTelemetryTracer()
+              .spanBuilder("com.google.cloud.bigquery.storage.v1.read.readRowsCallable")
```
shollyman, line 340: The behavior here is we'll get a single span that represents the lifetime of the server streaming rpc?  Can/should we add an attribute implicating the stream ID requested from the ReadRowsRequest?

shollyman (approved):

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/BigQueryReadClient.java
whuffman36: Stream ID seems like a good field to capture. I could be mistaken, but it seems like the ReadRowsRequest is handled by generated code and in this context we don't have access to the stream ID (ie can't put it in the attributes). Do you know of a way we can get the stream ID in the ReadClient or its enhanced stub @Neenu1995 ?

Neenu1995 (approved):

PR #2985, opened 2025-05-27
feat: provide append() methods that accept com.google.gson objects

agrawal-siddharth:


yirutang (commented):
There are too many format diffs, could we send in a change for pure format updates?

agrawal-siddharth:
I just realized I need to run the formatting using a new code formatter

yirutang (dismissed):

yirutang (approved):

PR #3086, opened 2025-09-04
feat(write): Support Arrow format for Write API

AlexZMLyu:

### DIFF
--- a/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/AppendFormats.java
+++ b/google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/AppendFormats.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigquery.storage.v1;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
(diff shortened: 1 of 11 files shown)


google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/AppendFormats.java
tswast, line 1: Design question: What are the benefits of providing these classes over using the autogenerated protobufs in the public interface instead? Are the oneof enums (https://protobuf.dev/reference/java/java-generated/#oneof-fields) not providing some abstraction that is needed?

Aside: It looks like https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#appendrowsrequest is out of date compared to  https://github.com/googleapis/googleapis/blob/8c79c6a821168af197149f86748a1cea408c2c99/google/cloud/bigquery/storage/v1/storage.proto#L424-L565 Have you talked with the tech writers about refreshing the docs?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
```diff
@@ -525,10 +648,29 @@ public String getStreamName() {
   /**
    * @return the passed in user schema.
    */
-  public ProtoSchema getProtoSchema() {
+  /** {@return the user provided schema in a general AppendRowsSchema} */
+  AppendRowsSchema getWriterSchema() {
```
tswast, line 652: Is it intended that this method wasn't marked as public?

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/AppendFormats.java
AlexZMLyu: The oneof enums can only deal with "rows" that has schema and data together. We need to deal with schema and data separately in different code paths.

Thanks for pointing out the missing arrow_rows in the doc and we do need to add it there.

google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1/StreamWriter.java
AlexZMLyu: Yes. This is intended.

We have 2 public methods below to get proto schema and arrow schema separately. We don't want user to  touch the package private AppendRowsSchema class.

tswast (approved):

